PR #29, opened 2025-04-24
Feat/embeddings

loicmathieu:


src/main/java/io/kestra/plugin/langchain4j/IngestDocument.java
```diff
@@ -0,0 +1,255 @@
...
+@ToString
+@EqualsAndHashCode
+@Getter
+@NoArgsConstructor
+@Schema(
+    title = "Index a document into an embedded store",
+    description = "Only text documents (TXT, HTML, Markdown) are supported for now."
+)
+@Plugin(
+    examples = {
```
fdelbrayelle, line 46: It could be great to add other examples for each kind of "from"

src/main/java/io/kestra/plugin/langchain4j/RAG.java
```diff
@@ -0,0 +1,149 @@
...
+                      apiKey: your_api_key
+                    embeddingStore:
+                      type: io.kestra.plugin.langchain4j.store.KvEmbeddingStore
+                    fromDocuments:
+                      - content: My name is Loïc
+                      - content: I live in Lille
+                      - content: My tailor is rich
+                  - id: rag
+                    type: io.kestra.plugin.langchain4j.RAG
+                    chatModelProvider:
```
fdelbrayelle, line 56: Use the same property name than `IndexDocument` for the harmony

src/main/java/io/kestra/plugin/langchain4j/RAG.java
```diff
@@ -0,0 +1,149 @@
...
+                id: rag
+                namespace: company.team
+
+                tasks:
+                  - id: ingest
+                    type: io.kestra.plugin.langchain4j.IndexDocument
+                    provider:
+                      type: io.kestra.plugin.langchain4j.gemini.GeminiModelProvider
+                      modelName: gemini-embedding-exp-03-07
+                      apiKey: your_api_key
```
fdelbrayelle, line 47: Maybe something like this so that in the doc we propagate the best practices to protect secrets:

```suggestion
                      apiKey: '{{ secret(...) }}'
```

src/test/java/io/kestra/plugin/langchain4j/RAGTest.java
```diff
@@ -0,0 +1,60 @@
...
+                    .modelName(new Property<>("{{ modelName }}"))
+                    .endpoint(new Property<>("{{ endpoint }}"))
+                    .build()
+            )
+            .embeddingStore(KvEmbeddingStore.builder().build())
+            .prompt(Property.of("Hello World"))
+            .build();
+
+        var ragOutput = rag.run(runContext);
+        assertThat(ragOutput.getCompletion()).isNotNull();
```
fdelbrayelle, line 58: Maybe check more about the content?

fdelbrayelle (approved):

src/main/java/io/kestra/plugin/langchain4j/RAG.java
loicmathieu: It's because there is two model provider, so if I use provider and embeddingModelProvider this is not clear.
In such cases we always prefer to be as specific as possible.
Anyway, we know that we will rename a lot of things later so I'll keep it as it for now.

src/test/java/io/kestra/plugin/langchain4j/RAGTest.java
loicmathieu: This is tricky as, well, you never know what will be the completion so asserting anything about the content may be flaky.

src/main/java/io/kestra/plugin/langchain4j/IngestDocument.java
loicmathieu: Examples will be totally rewritten by the content team later.

src/main/java/io/kestra/plugin/langchain4j/RAG.java
loicmathieu: Examples will be totally rewrtitten by the content team later.

src/main/java/io/kestra/plugin/langchain4j/IngestDocument.java
```diff
@@ -0,0 +1,255 @@
...
+        @Example(
+            full = true,
+            title = "Ingest documents into a KV embedding store.\\nWARNING: the KV embedding store is for quick prototyping only as it store the embedding vectors in a K/V store an load them all in memory.",
```
brian-mulier-p:
```suggestion
            title = "Ingest documents into a KV embedding store.\\nWARNING: the KV embedding store is for quick prototyping only as it stores the embedding vectors in a K/V Store and load them all in memory.",
```

src/main/java/io/kestra/plugin/langchain4j/IngestDocument.java
```diff
@@ -0,0 +1,255 @@
...
+    @Schema(
+        title = "Additional metadata that will be added to all ingested documents"
+    )
+    private Property<Map<String, String>> metadata;
+
+    @Schema(
+        title = "The document splitter"
+    )
+    @PluginProperty
+    private DocumentSplitter documentSplitter;
```
brian-mulier-p, line 116: Would be great to have a default for it, I see that you put RECURSIVE as default but then since we don't put a DocumentSplitter instance, user will need to provide some values.

Maybe add a default instance with some arbitrary values for sizes?

Btw could you add some javadoc on the splitter type enum values? (Even if it doesn't appear in the doc, at least it will help take over your code)

src/main/java/io/kestra/plugin/langchain4j/RAG.java
```diff
@@ -0,0 +1,149 @@
...
+    @Builder
+    @Getter
+    public static class ContentRetrieverConfiguration {
+        @Schema(title = "The maximum number of results from the embedding store.")
+        @Builder.Default
+        private Integer maxResults = 3;
+
+        @Schema(title = "The minimum score, ranging from 0 to 1 (inclusive). Only embeddings with a score >= minScore will be returned.")
+        @Builder.Default
+        private Double minScore = 0.0D;
```
brian-mulier-p, line 147: Maybe it's a too low default that defeats the purpose of using RAG? Or is it common thing to have a 0 score so that it includes everything at once?

src/test/java/io/kestra/plugin/langchain4j/IngestDocumentTest.java
```diff
@@ -0,0 +1,137 @@
...
+                    .build()
+            )
+            .embeddingStore(
+                KvEmbeddingStore.builder().build()
+            )
+            .fromDocuments(List.of(IngestDocument.InlineDocument.builder().content(Property.of("I'm Loïc")).build()))
+            .build();
+
+        IngestDocument.Output output = task.run(runContext);
+        assertThat(output.getIngestedDocuments()).isEqualTo(1);
```
brian-mulier-p, line 56: Can you assert the content of the KV?

src/test/java/io/kestra/plugin/langchain4j/IngestDocumentTest.java
```diff
@@ -0,0 +1,137 @@
...
+                    .build()
+            )
+            .embeddingStore(
+                KvEmbeddingStore.builder().build()
+            )
+            .fromInternalURIs(Property.of(List.of(uri.toString())))
+            .build();
+
+        IngestDocument.Output output = task.run(runContext);
+        assertThat(output.getIngestedDocuments()).isEqualTo(1);
```
brian-mulier-p, line 90: Same as above

src/test/java/io/kestra/plugin/langchain4j/IngestDocumentTest.java
```diff
@@ -0,0 +1,137 @@
...
+                    .build()
+            )
+            .embeddingStore(
+                KvEmbeddingStore.builder().build()
+            )
+            .fromPath(Property.of("ingest"))
+            .build();
+
+        IngestDocument.Output output = task.run(runContext);
+        assertThat(output.getIngestedDocuments()).isEqualTo(2);
```
brian-mulier-p, line 125: Same as above

src/test/java/io/kestra/plugin/langchain4j/IngestDocumentTest.java
```diff
@@ -0,0 +1,137 @@
...
+                    .build()
+            )
+            .embeddingStore(
+                KvEmbeddingStore.builder().build()
+            )
+            .fromExternalURLs(Property.of(List.of("https://dummyjson.com/products/1", "https://dummyjson.com/products/2")))
+            .build();
+
+        IngestDocument.Output output = task.run(runContext);
+        assertThat(output.getIngestedDocuments()).isEqualTo(2);
```
brian-mulier-p, line 155: Same as above, if unpredictable, just assert it's not null

src/test/java/io/kestra/plugin/langchain4j/RAGTest.java
brian-mulier-p: Maybe put a prompt like "What's my name?" and assert the response contains "Loïc"? I understand the flakiness but at least it proves that RAG is doing something

brian-mulier-p (approved):
LGTM if you fix the tiny comments I made

src/main/java/io/kestra/plugin/langchain4j/IngestDocument.java
loicmathieu: This is advise to split, but we cannot really define the best size by ourself.
RAG is hard, there is no good defaults, or I'm not expert enough to know some.

I'll add doc to the enum value, we definitely need a way to display that in our doc also, would be handy

src/main/java/io/kestra/plugin/langchain4j/RAG.java
loicmathieu: This is the default of langchain4j, as I have no idea what to put as default, I'll keep that to our product team.
Anyway, all tasks will be in beta so we can change that later.

src/test/java/io/kestra/plugin/langchain4j/IngestDocumentTest.java
loicmathieu: I cannot assert on the content of the embedding vector as it depends of the choice made by the embedding model.
But I'll check that we have one vector of embeddings

src/test/java/io/kestra/plugin/langchain4j/RAGTest.java
loicmathieu: You may put more trust in AI than me...
There was a test like that in the previous rag implementation, and it starts failing at some point...

loicmathieu: I'll do it anyway but don't complain to me later if it becomes flaky!

Issue #28, opened 2025-04-24
Add suppport for more language providers


loicmathieu:
### Feature description

- [ ] Anthropic
- [ ] MistralAI
- [ ] Deepseek

rkampani:
'd be happy to contribute to this. I assume we'll follow the existing structure and integrate each provider one by one with its corresponding library—such as the [langchain4j-anthropic API](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-anthropic/src/main) or other provider libraries. Let me know if it's okay for me to take a look.

loicmathieu:
Hi @rkampani it's exactly that ;) 
I assigned this issue to you, thanks for looking at it.
For any questions on contributing, please join our [Slack](https://kestra.io/slack) we have a dedicated contributors channel.

rkampani:
@loicmathieu : just wondering if you could help me out—I’m trying to build only the plugin module, but the build is failing because the ChatLanguageModel it depended on is no longer available in the upgraded (beta4) version. The issue seems to stem from a breaking change introduced in the [langchain4j/releases](https://github.com/langchain4j/langchain4j/releases). Do you have any suggestions on how to get the build working, or is it necessary to build the entire Kestra project in order to successfully build an individual module?

loicmathieu:
@rkampani I'm not sure I understand it, do you mean that Langchain4J version 1.0.0-beta4 is not working?
You can downgrade to 1.0.0-beta3 if needed, I'll sort this out later.

rkampani:
@loicmathieu : sure, I’ll try downgrading and see if that resolves the issue (JFYI: i'm trying plugin-langchain4j: master). Thanks for the quick response!beta-4 introduces breaking changes, and we’re currently relying on models that have been removed, as highlighted in this PR: [https://github.com/langchain4j/langchain4j/pull/2866](https://github.com/langchain4j/langchain4j/pull/2866).

Also, would it be possible to grant me access to trigger the build, so I can verify if the new changes compile & build successfully?

rkampani:
[...]

PR #46, opened 2025-05-12
Add support for more language providers #28

rkampani:

---

---

### DIFF
--- a/src/main/java/io/kestra/plugin/langchain4j/provider/AnthropicAI.java
+++ b/src/main/java/io/kestra/plugin/langchain4j/provider/AnthropicAI.java
@@ -1,0 +1,59 @@
+package io.kestra.plugin.langchain4j.provider;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import dev.langchain4j.model.anthropic.AnthropicChatModel;
+import dev.langchain4j.model.chat.ChatLanguageModel;
+import dev.langchain4j.model.embedding.EmbeddingModel;
+import dev.langchain4j.model.image.ImageModel;
+import io.kestra.core.exceptions.IllegalVariableEvaluationException;
+import io.kestra.core.models.annotations.Plugin;
+import io.kestra.core.models.property.Property;
+import io.kestra.core.runners.RunContext;
+import io.kestra.plugin.langchain4j.domain.ChatConfiguration;
+import io.kestra.plugin.langchain4j.domain.ModelProvider;
+import io.swagger.v3.oas.annotations.media.Schema;
+import jakarta.validation.constraints.NotNull;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.experimental.SuperBuilder;
+
+@Getter
+@SuperBuilder
+@NoArgsConstructor
+@AllArgsConstructor
+@Plugin(beta = true)
+@JsonDeserialize
+@Schema(
+    title = "Anthropic AI Model Provider"
+)
+public class AnthropicAI extends ModelProvider {
+    @Schema(title = "API Key")
+    @NotNull
+    private Property<String> apiKey;
+
+    @Override
+    public ChatLanguageModel chatLanguageModel(RunContext runContext, ChatConfiguration configuration) throws IllegalVariableEvaluationException {
+        return AnthropicChatModel.builder()
+            .modelName(runContext.render(this.getModelName()).as(String.class).orElseThrow())
+            .apiKey(runContext.render(this.apiKey).as(String.class).orElseThrow())
+            .temperature(runContext.render(configuration.getTemperature()).as(Double.class).orElse(null))
+            .topK(runContext.render(configuration.getTopK()).as(Integer.class).orElse(null))
+            .topP(runContext.render(configuration.getTopP()).as(Double.class).orElse(null))
+            .logRequests(true)
+            .logResponses(true)
+            .cacheSystemMessages(true)
+            .build();
+    }
+
+    @Override
+    public ImageModel imageModel(RunContext runContext) {
+        throw new UnsupportedOperationException("Anthropic didn't support image model");
+    }
+
+    @Override
+    public EmbeddingModel embeddingModel(RunContext runContext) throws IllegalVariableEvaluationException {
+        throw new UnsupportedOperationException("Anthropic doesn’t support embeddings");
+    }
+
+}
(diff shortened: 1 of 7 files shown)


src/main/java/io/kestra/plugin/langchain4j/provider/AnthropicAI.java
```diff
@@ -0,0 +1,59 @@
...
+    public ChatLanguageModel chatLanguageModel(RunContext runContext, ChatConfiguration configuration) throws IllegalVariableEvaluationException {
+        return AnthropicChatModel.builder()
+            .modelName(runContext.render(this.getModelName()).as(String.class).orElseThrow())
+            .apiKey(runContext.render(this.apiKey).as(String.class).orElseThrow())
+            .temperature(runContext.render(configuration.getTemperature()).as(Double.class).orElse(null))
+            .topK(runContext.render(configuration.getTopK()).as(Integer.class).orElse(null))
+            .topP(runContext.render(configuration.getTopP()).as(Double.class).orElse(null))
+            .logRequests(true)
+            .logResponses(true)
+            .cacheSystemMessages(true)
```
loicmathieu: I don't think we enable those properties in any other ChatModel, is there a reason why you did that?
For logRequest/logResponse, we usually avoid too much logging so better to turn them off (should be the default).

src/main/java/io/kestra/plugin/langchain4j/provider/DeepseekAI.java
```diff
@@ -0,0 +1,69 @@
...
+@Schema(
+    title = "Deepseek Model Provider"
+)
+public class DeepseekAI extends ModelProvider {
+    @Schema(title = "API Key")
+    @NotNull
+    private Property<String> apiKey;
+
+    @Schema(title = "API base URL")
+    private Property<String> baseUrl;
```
loicmathieu, line 38: Is it really needed? Is there multiple possible baseUrl for DeepSeek?
At least, set a default to the official DeepSeek API base URL

loicmathieu: as here it's not mandatory and the default would be to the official OpenAI API URL as it uses the OpenAiChatModel under the cover

src/main/java/io/kestra/plugin/langchain4j/provider/DeepseekAI.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    public ChatLanguageModel chatLanguageModel(RunContext runContext, ChatConfiguration configuration) throws IllegalVariableEvaluationException {
+        return OpenAiChatModel.builder()
```
loicmathieu, line 45: Can you add, as in the OpenAI provider a check that users didn't set topK as it's not supported by OpenAI?

```
if (configuration.getTopK() != null) {
            throw new IllegalArgumentException("OpenAI models didn't support topK");
        }
```

src/main/java/io/kestra/plugin/langchain4j/provider/MistralAI.java
```diff
@@ -0,0 +1,66 @@
...
+    @Schema(title = "API Key")
+    @NotNull
+    private Property<String> apiKey;
+
+    @Schema(title = "API base URL")
+    private Property<String> baseUrl;
+
+    @Override
+    public ChatLanguageModel chatLanguageModel(RunContext runContext, ChatConfiguration configuration) throws IllegalVariableEvaluationException {
+        return MistralAiChatModel.builder()
```
loicmathieu, line 45: If topK is not supported, can you add a check as for OpenAI and DeepSeek?

loicmathieu (changes requested):
Hi, thanks a lot for your contribution.

Just for us to know, did you test them or not? No problem if you didn't test them, we know that having API Key for all AI provider is not easy ;)

Ideally, you should also add example inside the tasks but I'll ask our tech writer to take care of it later.

src/main/java/io/kestra/plugin/langchain4j/provider/AnthropicAI.java
rkampani: I had enabled logging for my testing while wiring up the test cases — sorry, I completely forgot to remove it. I also added caching, since even Anthropic recommends it to help reprocess requests more efficiently.-so i added, will definitely remove it since we're not doing in any other models
https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching

rkampani:
> Hi, thanks a lot for your contribution.
> 
> Just for us to know, did you test them or not? No problem if you didn't test them, we know that having API Key for all AI provider is not easy ;)

[RK] I don’t have all the API keys needed to validate with every provider, so it would be great if someone from the team could help with that 

> Ideally, you should also add example inside the tasks but I'll ask our tech writer to take care of it later.

[RK] Oh, thank you! I’m new to this repo, so I’m not yet familiar with everything that’s needed to complete the full workflow. Really appreciate it!

src/main/java/io/kestra/plugin/langchain4j/provider/AnthropicAI.java
loicmathieu: Anyway, as we create a new model for each task execution I'm not sure caching will work.

loicmathieu:
@rkampani you can find here and example of a task inline documentation example: https://github.com/kestra-io/plugin-langchain4j/blob/master/src/main/java/io/kestra/plugin/langchain4j/ChatCompletion.java#L34

This will be automatically added to the documentation on both our website and the Kestra application UI.
But as I said, I can ask someone from our team to write those, if you're not familiar with Kestra it can be complex for you.

Your contribution is already of great value for us ;)

src/main/java/io/kestra/plugin/langchain4j/provider/DeepseekAI.java
rkampani: i believe OpenAI SDK defaults to OpenAI’s API URL (https://api.openai.com/v1) if no base_url is specified. This is because the SDK is primarily built for OpenAI’s services. It explicitly recommends setting the base_url to https://api.deepseek.com or https://api.deepseek.com/v1 when using the OpenAI SDK. [deepseek-documentation](https://api-docs.deepseek.com/). and 

Here is code L4j snippet

- [OpnAiChatModel](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java#L61) 
- [Default Urll](https://github.com/langchain4j/langchain4j/blob/04ef681053a38ee046f2ffb52d490d3acf7aed94/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/OpenAiUtils.java#L77)

rkampani:
Thank you for sharing the inline doc example. Most of the changes have been completed as suggested. Please re-review and let me know your thoughts on the baseUrl — DeepSeekAI ..I believe it recommended overwriting the base URL.

loicmathieu (approved):
LGTM thanks a lot for your contribution.

PR #47, opened 2025-05-13
feat(embeddings): add a Search task to query text into embeddings

### LINKED ISSUE #32: Add a SearchEmbedding task
### Feature description

TBD

yjrathod:
This change adds a new task feature: **SearchEmbedding** to the `langchain4j` plugin. (Closes #32)

Added three new Java classes under `io.kestra.plugin.langchain4j.embeddings`:

- `Search.java` – defines the task logic  
- `EmbeddingStoreProvider.java` – handles embedding store configuration  
- `ModelProvider.java` – handles embedding model configuration  

These changes enable searching embeddings using Hugging Face models and in-memory storage.


AcevedoR:
@loicmathieu assigning to you since you know the subject really well and requested it 😄

src/main/java/io/kestra/plugin/langchain4j/embeddings/EmbeddingStoreProvider.java
loicmathieu, line 1: This interface is not needed, please use the existing abstract class `io.kestra.plugin.langchain4j.domain.EmbeddingStoreProvider` instead.

src/main/java/io/kestra/plugin/langchain4j/embeddings/ModelProvider.java
```diff
@@ -0,0 +1,7 @@
+package io.kestra.plugin.langchain4j.models;
+
+import dev.langchain4j.model.embedding.EmbeddingModel;
+
+public interface ModelProvider {
```
loicmathieu: This interface is not needed, please use the existing `io.kestra.plugin.langchain4j.domain.ModelProvider` instead.

loicmathieu (changes requested):
Can you provide tests?
At minima based on OLlama, you can check the RAG test for how to build such test.

fdelbrayelle:
Hi @yjrathod :wave: 
Could you give us a feedback regarding the requested changes please?
Once they will be taken into account we could merge this PR :rocket: 
Thank you!

src/main/java/io/kestra/plugin/langchain4j/rag/Search.java
```diff
@@ -0,0 +1,122 @@
...
+@SuperBuilder
+@Getter
+@NoArgsConstructor
+@ToString
+@EqualsAndHashCode
+@Plugin(
+    examples = {
+        @Example(
+            full = true,
+            title = "Search using embeddings from a query against an embedding store.",
```
loicmathieu:
```suggestion
            title = "Make a search query against an embedding store.",
```

src/main/java/io/kestra/plugin/langchain4j/rag/Search.java
```diff
@@ -0,0 +1,122 @@
...
+                    embeddings:
+                      type: io.kestra.plugin.langchain4j.embeddings.KestraKVStore
+                    query: "Feature Highlights"
+                """
+        ),
+    },
+    beta = true
+)
+@Schema(
+    title = "Search text into embeddings",
```
loicmathieu:
```suggestion
    title = "Search from an embedding store",
```

src/main/java/io/kestra/plugin/langchain4j/rag/Search.java
```diff
@@ -0,0 +1,122 @@
...
+    @NotNull
+    private Property<String> query;
+
+    @Schema(title = "Maximum number of results to return")
+    @Builder.Default
+    private Property<Integer> maxResults = Property.ofValue(5);
+
+    @Schema(title = "Minimum similarity score")
+    @Builder.Default
+    private Property<Double> minScore = Property.ofValue(0.8);
```
loicmathieu: This is highly opinionated.
The default in Langchain4J is 0, we may use that.
Or we may use a less high value like 0.5

src/main/java/io/kestra/plugin/langchain4j/rag/Search.java
```diff
@@ -0,0 +1,122 @@
...
+)
+public class Search extends Task implements RunnableTask<Search.Output> {
+
+    @Schema(title = "Query string to search for")
+    @NotNull
+    private Property<String> query;
+
+    @Schema(title = "Maximum number of results to return")
+    @Builder.Default
+    private Property<Integer> maxResults = Property.ofValue(5);
```
loicmathieu: I alwyas prefer to made mandatory properties that have default (and use orElseThrow when retrieving)

src/main/java/io/kestra/plugin/langchain4j/rag/Search.java
loicmathieu: I alwyas prefer to made mandatory properties that have default (and use orElseThrow when retrieving)

src/main/java/io/kestra/plugin/langchain4j/rag/Search.java
```diff
@@ -0,0 +1,122 @@
...
+            .toList();
+
+        return Output.builder()
+            .results(results)
+            .build();
+    }
+
+    @Builder
+    @Getter
+    public static class Output implements io.kestra.core.models.tasks.Output {
```
loicmathieu, line 199: Can you have a `size` (or count) attribute with then number of results?
We usually add that, it's easier to use than requesting in Pebble the size of the result list.

loicmathieu (approved):
LGTM, some suggestion that you can apply before merging but no need to ask for a second review

src/main/java/io/kestra/plugin/langchain4j/rag/Search.java
fdelbrayelle: Yes and I'll add a fetch type in addition

PR #48, opened 2025-05-16
chore: renaming and doc

loicmathieu:


loicmathieu:
Merging without approval as it's only doc and renaming

PR #52, opened 2025-05-23
Feat/remove text completion

loicmathieu:


Malaydewangan09 (approved):
Can we also now try to use new `Property.ofValue()` since the `Property.of()` is deprecated now ?

loicmathieu:
@Malaydewangan09 yeah, well, I didn't rebuild locally so I haven't the method available when I did this PR. That's why I didn't use it.

PR #57, opened 2025-06-02
feat(embeddings): introduce new embedding stores

### LINKED ISSUE #49: Add more Embedding Stores for RAG use case in Langchain4j
### Feature description

Current state of RAG plugin: `io.kestra.plugin.langchain4j.rag`

We have those 3 embedding stores io.kestra.plugin.langchain4j.embeddings:

1. Elasticsearch
2. KestraKVStore
3. PGVector

Below is a list of other embedding stores (mainly vector databases) that we want to consider:
- [x] Chroma
- [x] Qdrant
- [x] Pinecone
- [x] Weaviate
- [x] FAISS
- [x] Milvus

Optional extra:
- [x] AstraDB
- [x] Redis
- [x] Clickhouse
- [x] Databricks
- [x] MongoDB Atlas

fdelbrayelle:

---

---

### Setup Instructions

fixes https://github.com/kestra-io/plugin-langchain4j/issues/49

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -72,6 +72,7 @@ dependencies {
     implementation "dev.langchain4j:langchain4j-pinecone"
     implementation "dev.langchain4j:langchain4j-pgvector"
     implementation "dev.langchain4j:langchain4j-qdrant"
+    implementation "dev.langchain4j:langchain4j-weaviate"
 
     // tools
     implementation "dev.langchain4j:langchain4j-web-search-engine-google-custom"
--- a/build.gradle
+++ b/build.gradle
@@ -68,7 +68,6 @@ dependencies {
     // Langchain4j embeddings stores
     implementation "dev.langchain4j:langchain4j-chroma"
     implementation "dev.langchain4j:langchain4j-elasticsearch"
-    implementation "dev.langchain4j:langchain4j-embeddings-all-minilm-l6-v2"
     implementation "dev.langchain4j:langchain4j-mongodb-atlas"
     implementation "dev.langchain4j:langchain4j-milvus"
     implementation "dev.langchain4j:langchain4j-pinecone"
@@ -82,6 +81,8 @@ dependencies {
     implementation "dev.langchain4j:langchain4j-mcp"
 
     implementation "com.azure:azure-identity" // For Azure OpenAI
+
+    implementation 'org.awaitility:awaitility:4.2.0'
 }
 
 
@@ -135,6 +136,7 @@ dependencies {
     testImplementation 'org.testcontainers:elasticsearch:1.21.1'
 
     // Embeddings stores
+    testImplementation "dev.langchain4j:langchain4j-embeddings-all-minilm-l6-v2"
     testImplementation 'org.testcontainers:chromadb:1.19.6'
     testImplementation 'org.testcontainers:milvus:1.19.6'
     testImplementation 'org.testcontainers:mongodb:1.20.4'
--- a/build.gradle
+++ b/build.gradle
@@ -66,7 +66,9 @@ dependencies {
     implementation "dev.langchain4j:langchain4j-vertex-ai-gemini"
 
     // Langchain4j embeddings stores
+    implementation "dev.langchain4j:langchain4j-chroma"
     implementation "dev.langchain4j:langchain4j-elasticsearch"
+    implementation "dev.langchain4j:langchain4j-embeddings-all-minilm-l6-v2" // for Chroma
     implementation "dev.langchain4j:langchain4j-pgvector"
     implementation "dev.langchain4j:langchain4j-qdrant"
 
@@ -127,6 +129,9 @@ dependencies {
 
     testImplementation 'org.testcontainers:ollama:1.21.1'
     testImplementation 'org.testcontainers:elasticsearch:1.21.1'
+
+    // Chroma
+    testImplementation 'org.testcontainers:chromadb:1.19.6'
 }
 
 /**********************************************************************************************************************\
--- a/build.gradle
+++ b/build.gradle
@@ -69,6 +69,7 @@ dependencies {
     implementation "dev.langchain4j:langchain4j-chroma"
     implementation "dev.langchain4j:langchain4j-elasticsearch"
     implementation "dev.langchain4j:langchain4j-embeddings-all-minilm-l6-v2" // for Chroma
+    implementation "dev.langchain4j:langchain4j-pinecone"
     implementation "dev.langchain4j:langchain4j-pgvector"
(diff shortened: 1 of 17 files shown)


build.gradle
```diff
@@ -53,19 +53,29 @@ dependencies {
...
     implementation "dev.langchain4j:langchain4j-ollama"
     implementation "dev.langchain4j:langchain4j-open-ai"
-    implementation "dev.langchain4j:langchain4j-pgvector"
     implementation "dev.langchain4j:langchain4j-vertex-ai"
     implementation "dev.langchain4j:langchain4j-vertex-ai-gemini"
 
+    // Langchain4j embeddings stores
+    implementation "dev.langchain4j:langchain4j-chroma"
+    implementation "dev.langchain4j:langchain4j-elasticsearch"
+    implementation "dev.langchain4j:langchain4j-embeddings-all-minilm-l6-v2"
```
loicmathieu: What is it used for? It didn't seems like an embedding store on its own?

src/main/java/io/kestra/plugin/langchain4j/embeddings/Chroma.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @NotNull
+    @Schema(title = "The collection name")
+    private Property<String> collectionName;
+
+    @Schema(title = "Whether the requests are logged or not.")
+    private Property<Boolean> logRequests;
+
+    @Schema(title = "Whether the responses are logged or not.")
+    private Property<Boolean> logResponses;
```
loicmathieu: We didn't expose those in any model provider or embedding store for now so if you're doing that, better to add it everywhere or nowhere.

src/main/java/io/kestra/plugin/langchain4j/embeddings/MongoDBAtlas.java
```diff
@@ -0,0 +1,132 @@
...
+            .build();
+
+        if (renderedCreateIndex) {
+            // Creating a vector search index can take up to a minute, so this delay allows the index to become queryable
+            try {
+                Thread.sleep(15000);
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
+            }
+        }
```
loicmathieu: This is really not a good idea.
Moreover, if the index already exist it would not take a lot of time.

The best would be to use awaitability and wait for the index to be ready by querying it or any other sort of query that make sure the index is ready.

src/main/java/io/kestra/plugin/langchain4j/embeddings/MongoDBAtlas.java
```diff
@@ -0,0 +1,132 @@
...
+        // Format: mongodb+srv://[username:password@]host[/[database][?options]]
+
+        var username = runContext.render(this.username).as(String.class).orElse(null);
+        var password = runContext.render(this.password).as(String.class).orElse(null);
+        var host = runContext.render(this.host).as(String.class).orElseThrow();
+        var database = runContext.render(this.database).as(String.class).orElseThrow();
+        var options = runContext.render(this.options).asMap(String.class, Object.class);
+
+        boolean isAtlas = host.endsWith(".mongodb.net");
+        var scheme = isAtlas ? "mongodb+srv" : "mongodb"; // mongodb here is just for test purposes
```
loicmathieu: I think `mongo+svr` can be used for more than MongoDB atlas. The best would be to let the user configure the URI by itself instead of crafting it for it as there are a lot of different URI possible.

build.gradle
loicmathieu: It seems to be used only for test, if so, add it as test dependency.

loicmathieu (changes requested):

src/main/java/io/kestra/plugin/langchain4j/embeddings/Chroma.java
fdelbrayelle: They weren't available for other new stores I added but I can remove them here it's fine :+1:

build.gradle
fdelbrayelle: Yes it's for test purposes so `testImplementation` you're right!

src/main/java/io/kestra/plugin/langchain4j/embeddings/MongoDBAtlas.java
fdelbrayelle: Yes at first I thought to let the user set the full uri but it contains info I need elsewhere such as the database. And plus I applied the "Never trust user!" rule who can put a lot of error in such a connection string, what do you think?

src/main/java/io/kestra/plugin/langchain4j/rag/IngestDocument.java
```diff
@@ -106,14 +106,133 @@
...
+                tasks:
+                  - id: ingest
+                    type: io.kestra.plugin.langchain4j.rag.IngestDocument
+                    provider:
+                      type: io.kestra.plugin.langchain4j.provider.GoogleGemini
+                      modelName: gemini-embedding-exp-03-07
+                      apiKey: my_api_key
+                    embeddings:
+                      type: io.kestra.plugin.langchain4j.embeddings.Milvus
+                      milvus: "{{ secret('MILVUS_TOKEN') }}"
```
Malaydewangan09:
```suggestion
                      token: "{{ secret('MILVUS_TOKEN') }}"
```

src/main/java/io/kestra/plugin/langchain4j/embeddings/MongoDBAtlas.java
fdelbrayelle: Done

Malaydewangan09 (approved):

src/main/java/io/kestra/plugin/langchain4j/embeddings/MongoDBAtlas.java
```diff
@@ -0,0 +1,151 @@
...
+                        .metadataFieldNames(new HashSet<>(runContext.render(metadataFieldNames).asList(String.class)))
+                        .build() : null
+            )
+            .build();
+
+        if (renderedCreateIndex) {
+            // Creating a vector search index can take up to a minute, so this delay allows the index to become queryable
+            Awaitility.await()
+                .atMost(Duration.ofMinutes(1))
+                .pollInterval(Duration.ofSeconds(3))
```
loicmathieu: I think by default it will do an exponential retry which would be better that using a fixed interval.
Can you remove pollInterval? Or set it to a lower value as 1s.

loicmathieu (approved):
LGTM, just a small remark on poll Interval which could be removed IMHO.

Issue #35, opened 2025-04-28
Add support for tools


loicmathieu:
### Feature description

Needs investiguation first as I'm not sure if we should provide it only in rag.Chat or in other tasks also.

As tools can be of any sort, a good start is to implements:

- [x] Web Search: https://docs.langchain4j.dev/category/web-search-engines
- [x] MCP Client: https://docs.langchain4j.dev/tutorials/mcp

rkampani:
I'd love to contribute to this. Could you please clarify the expectations?

loicmathieu:
As I said, this needs further investiguation but inside a RAG Chat, you can specify tools for the LLM to use.
At least two seems interesting: Web Search and MCP client.

The first step would be to prototype to see if it's easy to integrat into the existing rag.Chat task, and the increase in size that this would imply as this plugin is already big.

rkampani:
Great, that helps clarify things. I'll work on a POC and push the branch—then we can review to see if it meets the requirements and go from. there...

loicmathieu:
Yes, open a draft PR with your PoC and we'll discuss together to see what we want here.

rkampani:
I wasn't entirely sure if I fully understood the requirements beyond the integration with a web search engine, so did a [PoC](https://github.com/kestra-io/plugin-langchain4j/pull/62)  using Google and Tavily for testing. I wanted to share , and check if i correctly interpreting the expectations. for https://github.com/kestra-io/plugin-langchain4j/pull/62...  Please feel free to close the PR if it doesn’t align with what’s needed.

loicmathieu:
@rkampani interesting, it seems to be in the right direction, I'll comment in the PR.

loicmathieu:
Done in https://github.com/kestra-io/plugin-langchain4j/pull/62 and https://github.com/kestra-io/plugin-langchain4j/pull/65

PR #62, opened 2025-06-03
Add support for tools #35

rkampani:

---

---


loicmathieu:
Hi, thanks a lot for your PR.
I think it misses some part to be fully workable inside a real Kestra instance but we can provide it as it's a little too technical so let's focus on an implementation that work as a unit test first.

I think the best would be to have a tool abstraction instead of just a websearch abstraction, you can go from one to another via ` WebSearchTool webTool = WebSearchTool.from(searchEngine);` so when we add later MCP or even script we didn't add a new abstraction.

One important question is the size added by the new libraries.
Can you build the plugin without and with your changes to see how much increase in size?
To build the plugin run `gradle shadowJar` at the root of the repository, the plugin will be built inside `build/lib`.

rkampani:
- I’ve updated the code as suggested, using the tool as an abstraction layer.
- The query is now routed to both the Tavily and Google retrievers based on user selection, with `maxResults` defaulting to 3 (overridable by the user). This behavior has been validated through JUnit tests, which are conditionally enabled based on the relevant environment properties.
- The `DefaultRetrievalAugmentor` aggregates content from both retrievers and injects it into the LLM's prompt. I haven't customized this behavior with a specific `ContentAggregator` (e.g., `ReRankingContentAggregator`) yet, but incorporating re-ranking based on user input could be a enhancement for the future.
- The size increased from ~150.8 MB to ~150.3 MB due to these additions. 
plugin-langchain4j-afterChange
plugin-langchain4j-beforeChange
Screenshot 2025-06-04 at 4 55 45 PM

loicmathieu:
Hi,
Thanks a lot, this is amazing!
I'll do some adjustment to this PR for making it run properly inside a Kestra instance  then I'll merge it.

loicmathieu (approved):
LGTM, thanks a lot for your contribution

src/main/java/io/kestra/plugin/langchain4j/tool/GoogleCustomWebSearchTool.java
```diff
@@ -0,0 +1,55 @@
...
+@SuperBuilder
+@NoArgsConstructor
+@AllArgsConstructor
+@Plugin(beta = true)
+@JsonDeserialize
+@Schema(
+    title = "WebSearch tool for Google Custom Search"
+)
+public class GoogleCustomWebSearchTool extends ToolProvider {
+    @Schema(title = "API Key")
```
fdelbrayelle, line 31: Should be CSI (or something else here) @loicmathieu

src/main/java/io/kestra/plugin/langchain4j/tool/GoogleCustomWebSearchTool.java
```diff
@@ -0,0 +1,55 @@
...
+    private Property<String> csi;
+
+    @Schema(title = "API Key")
+    @NotNull
+    private Property<String> apiKey;
+
+    @Schema(title = "Maximum number of results to return")
+    @NotNull
+    @Builder.Default
+    private Property<Integer> maxResults = Property.ofValue(3);
```
fdelbrayelle, line 42: @loicmathieu  We converged to avoid default values for such properties

src/test/java/io/kestra/plugin/langchain4j/rag/ChatCompletionTest.java
```diff
@@ -17,6 +20,10 @@
 
 @KestraTest
 class ChatCompletionTest extends ContainerTest {
+    private final String GOOGLE_API_KEY = System.getenv("GOOGLE_API_KEY");
+    private final String GOOGLE_CSI = System.getenv("GOOGLE_CSI_KEY");
```
fdelbrayelle, line 24: GOOGLE_CSI_KEY

src/main/java/io/kestra/plugin/langchain4j/tool/GoogleCustomWebSearchTool.java
loicmathieu: I was not part of this decision as we previously recommend to do such so that it's properly validated and documented.

PR #64, opened 2025-06-05
feat: content provider and tools

loicmathieu:
Allow using both content providers and tools in RAG. 
Provides tools for standard ChatCompletion.
Allow using RAG without embeddings if there is a content provider.

### DIFF
--- a/src/main/java/io/kestra/plugin/langchain4j/ChatCompletion.java
+++ b/src/main/java/io/kestra/plugin/langchain4j/ChatCompletion.java
@@ -1,9 +1,18 @@
 package io.kestra.plugin.langchain4j;
 
+import dev.langchain4j.agent.tool.ToolSpecification;
+import dev.langchain4j.agent.tool.ToolSpecifications;
 import dev.langchain4j.data.message.AiMessage;
 import dev.langchain4j.data.message.SystemMessage;
+import dev.langchain4j.data.message.ToolExecutionResultMessage;
 import dev.langchain4j.data.message.UserMessage;
 import dev.langchain4j.model.chat.ChatModel;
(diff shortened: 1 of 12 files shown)


src/main/java/io/kestra/plugin/langchain4j/retriever/GoogleCustomWebSearchContentRetriever.java
```diff
@@ -0,0 +1,55 @@
...
+    private Property<String> csi;
+
+    @Schema(title = "API Key")
+    @NotNull
+    private Property<String> apiKey;
+
+    @Schema(title = "Maximum number of results to return")
+    @NotNull
+    @Builder.Default
+    private Property<Integer> maxResults = Property.ofValue(3);
```
fdelbrayelle: As we said for another properties in this plugin, let's not put a default value here since in addition it's already not null

src/main/java/io/kestra/plugin/langchain4j/retriever/TavilyWebSearchContentRetriever.java
```diff
@@ -0,0 +1,50 @@
...
+)
+public class TavilyWebSearchContentRetriever extends ContentRetrieverProvider {
+    @Schema(title = "API Key")
+    @NotNull
+    private Property<String> apiKey;
+
+    @Schema(title = "Maximum number of results to return")
+    @NotNull
+    @Builder.Default
+    private Property<Integer> maxResults = Property.ofValue(3);
```
fdelbrayelle: Same comment than previously for the default value

fdelbrayelle (approved):

PR #74, opened 2025-06-20
fix: repackage completion tasks into their own package

loicmathieu:
The root package cannot be used as a subgroup package so when using subgroup in a plugin all plugin must be in a dedicated package.


loicmathieu:
I added aliases even if it's still in beta because we provide it in 0.22 and have already a lot of contents out with it so I prefer to avoid all those contents to no longer work.
It will be backported to 0.23 and 0.22.

fdelbrayelle (approved):

PR #87, opened 2025-06-27
feat: AIAgent task with a memory

loicmathieu:


loicmathieu:
The AiAgent task itself will be redesigned later.

KestraTaskCalling tool will be provided in the following PR: https://github.com/kestra-io/plugin-ai/pull/105

PR #105, opened 2025-07-31
feat: Kestra task calling as a tool

### LINKED ISSUE #83: Evaluate how to have "Kestra Tasks as Tool"
It would be really nice to be able to call Kestra task parts of the tools of a given agent.

After discussion with @loicmathieu and @fdelbrayelle, this need more investigation on the technical side. Acc. to Loic there's a potential quick win.

loicmathieu:
Part-of: https://github.com/kestra-io/kestra-ee/issues/4582


src/main/java/io/kestra/plugin/ai/memory/KestraKVStore.java
```diff
@@ -90,9 +90,9 @@
                 """
         ),
     },
-    aliases = "io.kestra.plugin.langchain4j.memory.KestraKVMemory"
+    aliases = { "io.kestra.plugin.langchain4j.memory.KestraKVMemory", "io.kestra.plugin.ai.memory.KestraKVMemory" }
```
fdelbrayelle, line 93: Don't we want to remove the aliases for 1.0?

src/main/java/io/kestra/plugin/ai/tool/KestraTaskCalling.java
```diff
@@ -0,0 +1,248 @@
...
+
+    @Override
+    public Map<ToolSpecification, ToolExecutor> tool(RunContext runContext) throws IllegalVariableEvaluationException {
+        var defaultRunContext = (DefaultRunContext) runContext;
+        var jsonSchemaGenerator = defaultRunContext.getApplicationContext().getBean(JsonSchemaGenerator.class);
+
+        Map<ToolSpecification, ToolExecutor> tools = new HashMap<>();
+        for (Task task : tasks) {
+            // First, check that the task is a runnable task.
+            if (!(task instanceof RunnableTask)) {
```
fdelbrayelle, line 87: We don't want to have flowable tasks to do more complex things?

src/main/java/io/kestra/plugin/ai/memory/KestraKVStore.java
loicmathieu: Usually we don't add aliases for beta tasks, but when we add an alias we never remove it as it comes with (almost) not cost.
Now, there is aliases in this plugin even if in beta as we backport it down to 0.22 and had a lot of influencer/marketing on it so we decide to add aliases.

src/main/java/io/kestra/plugin/ai/tool/KestraTaskCalling.java
loicmathieu: Not possible with the current running model, we directly run the runnable task, we can't do that for flowable tasks

loicmathieu: Flowable tasks are not "runned" they are part of the executor which treat them very specifically.

fdelbrayelle (approved):

PR #107, opened 2025-08-07
feat: Call a Kestra flow as a tool

### LINKED ISSUE #108: Add a Kestra Flow Calling tool
### Feature description

Add a tool to call a Kestra Flow from an LLM.

This tool should allow both control by the user or the LLM, this means that inputs and labels you be settable inside the tool definition or dynamically by the LLM at generation time.

The namespace and id of the flow must be defined inside the tool, this tool is to call a defined flow not any existing flow which would have too high risk of hallucination.

loicmathieu:
Call a Kestra flow as a tool.

Closes #108

As discussed internally, for tools, we will use a pattern of: "being able to constrain the IA for more predictability without preventing dynamic usage" which means we restrict the IA for some choice (which flow to use) and we allow the user to optionally set some for better predictability or to let the IA set it for more dynamicity (labels, inputs, scheduleDate).

### DIFF
--- a/src/main/java/io/kestra/plugin/ai/tool/KestraFlowCalling.java
+++ b/src/main/java/io/kestra/plugin/ai/tool/KestraFlowCalling.java
@@ -1,0 +1,341 @@
+package io.kestra.plugin.ai.tool;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import dev.langchain4j.agent.tool.ToolExecutionRequest;
+import dev.langchain4j.agent.tool.ToolSpecification;
+import dev.langchain4j.model.chat.request.json.JsonArraySchema;
+import dev.langchain4j.model.chat.request.json.JsonObjectSchema;
+import dev.langchain4j.model.chat.request.json.JsonStringSchema;
+import dev.langchain4j.service.tool.ToolExecutor;
+import io.kestra.core.exceptions.IllegalVariableEvaluationException;
+import io.kestra.core.models.Label;
+import io.kestra.core.models.annotations.Example;
+import io.kestra.core.models.annotations.Plugin;
+import io.kestra.core.models.annotations.PluginProperty;
+import io.kestra.core.models.executions.Execution;
+import io.kestra.core.models.flows.FlowInterface;
+import io.kestra.core.models.property.Property;
+import io.kestra.core.queues.QueueFactoryInterface;
+import io.kestra.core.queues.QueueInterface;
+import io.kestra.core.runners.DefaultRunContext;
+import io.kestra.core.runners.FlowMetaStoreInterface;
+import io.kestra.core.runners.RunContext;
+import io.kestra.core.serializers.JacksonMapper;
+import io.kestra.core.serializers.ListOrMapOfLabelDeserializer;
+import io.kestra.core.serializers.ListOrMapOfLabelSerializer;
+import io.kestra.core.utils.IdUtils;
+import io.kestra.core.utils.ListUtils;
+import io.kestra.core.utils.MapUtils;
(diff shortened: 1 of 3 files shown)


fdelbrayelle:
This closes https://github.com/kestra-io/kestra-ee/issues/4582 ?

loicmathieu:
@fdelbrayelle no, it's the one for Flow not Task, but I cannot find any existing issue about it, I'll create one.

src/main/java/io/kestra/plugin/ai/tool/KestraFlowCalling.java
```diff
@@ -0,0 +1,373 @@
...
+                        throw new IllegalArgumentException("You need to provide an input with the id '" + input.getId() + "'.");
+                    }
+                });
+
+                var execution = Execution.newExecution(flowInterface, (f, e) -> finalInputs, finalLabels, scheduledDate);
+
+                var executionQueue = (QueueInterface<Execution>) runContext.getApplicationContext().getBean(QueueInterface.class, Qualifiers.byName(QueueFactoryInterface.EXECUTION_NAMED));
+                executionQueue.emit(execution);
+
+                return "Success";
```
fdelbrayelle, line 335: Not exploited anywhere?

src/main/java/io/kestra/plugin/ai/tool/KestraFlowCalling.java
```diff
@@ -0,0 +1,373 @@
...
+import java.util.*;
+import java.util.stream.Collectors;
+
+import static io.kestra.core.utils.Rethrow.throwConsumer;
+
+@Getter
+@SuperBuilder
+@NoArgsConstructor
+@Plugin(
+    beta = true,
```
fdelbrayelle, line 50: Didn't we say it's not beta anymore?

src/main/java/io/kestra/plugin/ai/tool/KestraFlowCalling.java
loicmathieu: No, it just tells the LLM that the tool has been called successful.
If there is an error in the tool, the LLM will call the tool again, maybe with different parameters (if the error suggest it's a parameter error).

src/main/java/io/kestra/plugin/ai/tool/KestraFlowCalling.java
loicmathieu: Yes but I prefer that all plugins would be switched at the same time.
Moreover, we may want to set as beta all the new agent related plugins if we're not 100% sure as those would be tricky to design (this one is the simplest).

src/main/java/io/kestra/plugin/ai/tool/KestraFlowCalling.java
```diff
@@ -0,0 +1,373 @@
...
+                      - id: first
+                        type: io.kestra.plugin.ai.completion.ChatCompletion
+                        provider:
+                          type: io.kestra.plugin.ai.provider.GoogleGemini
+                          modelName: gemini-2.5-flash
+                          apiKey: "{{ secret('GOOGLE_API_KEY') }}"
+                        tools:
+                          - type: io.kestra.plugin.ai.tool.KestraFlowCalling
+                            namespace: company.team
+                            flowId: hello-world
```
fdelbrayelle, line 76: Shouldn't you add the `hello-world` flow below?

fdelbrayelle (approved):

src/main/java/io/kestra/plugin/ai/tool/KestraFlowCalling.java
loicmathieu: It can be any flow, it didn't provide any added value here. KISS!

I decide to add the flow in the other example to show inputs/labels usage so it's a more involved example.

PR #124, opened 2025-08-19
Fix/memory drop

loicmathieu:


src/main/java/io/kestra/plugin/ai/memory/Redis.java
```diff
@@ -117,15 +118,20 @@ public ChatMemory chatMemory(RunContext runContext) throws IllegalVariableEvalua
...
         this.chatMemory = MessageWindowChatMemory.withMaxMessages(runContext.render(this.getMessages()).as(Integer.class).orElseThrow());
         var key = runContext.render(this.getMemoryId()).as(String.class).orElseThrow();
 
         try (var jedis = new Jedis(rHost, rPort)) {
             var json = jedis.get(key);
             if (json != null) {
-                var messages = ChatMessageDeserializer.messagesFromJson(json);
-                messages.forEach(chatMemory::add);
+                if (rDrop == Drop.BEFORE_EXECUTION) {
+                    jedis.del(key);
```
fdelbrayelle, line 130: Shouldn't we call `chatMemory::add` and the rest after this then?

loicmathieu: No, this is the trick: dropping before the excution means we didn't use the message in the store and we drop what's in it.

fdelbrayelle (approved):

PR #126, opened 2025-08-20
Feat/improvemebts

loicmathieu:


src/main/java/io/kestra/plugin/ai/tool/internal/JsonObjectSchemaTranslator.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    @SuppressWarnings("unchecked")
+    public static JsonObjectSchema fromOpenAPISchema(Map<String, Object> schema, String description) {
+        var definitions = mapDefinitions((Map<String, Object>) schema.get("$defs"));
+        var properties = mapProperties((Map<String, Object>) schema.get("properties"));
+
+        // some LLM didn't support definitions, so we will remove them and replace them by their schema.
+        definitions = replaceDefinitions(definitions, definitions);
+        properties = replaceDefinitions(properties, definitions);
```
fdelbrayelle, line 22: Are you sure of the params here?

fdelbrayelle (approved):

src/main/java/io/kestra/plugin/ai/tool/internal/JsonObjectSchemaTranslator.java
loicmathieu: Yes, a definition can itself have definitions so we must replace definitions inside definitions...

PR #129, opened 2025-08-21
fix(tests): ensure all models are pulled before tests

fdelbrayelle:

---

---


loicmathieu (approved):
I'm approving but overall, the timeout are huge and manual retry and thread.sleep is never a good idea. Would be better to use Awaitility for that.

fdelbrayelle:
@loicmathieu Yes I'll lower the timeout and use Awaitility you're right :+1:

PR #156, opened 2025-09-03
fix(provider-openai): default creds handling

Malaydewangan09:
Since for `DefaultAzureCredentialBuilder` also, tenantId is also not required, so no need to check for null.

> Sets the tenant id of the user to authenticate through the {'@'link DefaultAzureCredential}. If unset, the value
>  in the AZURE_TENANT_ID environment variable will be used. If neither is set, the default is null
>  and will authenticate users to their default tenant.

https://kestra-io.slack.com/archives/C07KA8R1K0F/p1756920900243799


brian-mulier-p (approved):
LGTM

fdelbrayelle (approved):

PR #170, opened 2025-09-23
feat: agent as tool

loicmathieu:
Part-of: https://github.com/kestra-io/kestra-ee/issues/4786

### DIFF
--- a/src/main/java/io/kestra/plugin/ai/tool/AIAgent.java
+++ b/src/main/java/io/kestra/plugin/ai/tool/AIAgent.java
@@ -1,0 +1,214 @@
+package io.kestra.plugin.ai.tool;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import dev.langchain4j.agent.tool.ToolExecutionRequest;
+import dev.langchain4j.agent.tool.ToolSpecification;
+import dev.langchain4j.exception.ToolArgumentsException;
+import dev.langchain4j.exception.ToolExecutionException;
+import dev.langchain4j.model.chat.request.json.JsonObjectSchema;
+import dev.langchain4j.rag.DefaultRetrievalAugmentor;
+import dev.langchain4j.rag.content.retriever.ContentRetriever;
+import dev.langchain4j.rag.query.router.DefaultQueryRouter;
+import dev.langchain4j.rag.query.router.QueryRouter;
+import dev.langchain4j.service.AiServices;
+import dev.langchain4j.service.Result;
+import dev.langchain4j.service.tool.ToolExecutor;
+import io.kestra.core.exceptions.IllegalVariableEvaluationException;
+import io.kestra.core.models.annotations.Example;
+import io.kestra.core.models.annotations.Plugin;
+import io.kestra.core.models.annotations.PluginProperty;
+import io.kestra.core.models.property.Property;
+import io.kestra.core.runners.RunContext;
+import io.kestra.core.serializers.JacksonMapper;
+import io.kestra.core.utils.ListUtils;
+import io.kestra.plugin.ai.AIUtils;
+import io.kestra.plugin.ai.domain.*;
+import io.swagger.v3.oas.annotations.media.Schema;
+import jakarta.validation.constraints.NotNull;
+import lombok.AccessLevel;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.experimental.SuperBuilder;
+
+import java.util.List;
+import java.util.Map;
+
+import static io.kestra.core.utils.Rethrow.throwFunction;
+
+@Getter
+@SuperBuilder
+@NoArgsConstructor
+@Plugin(
+    examples =  {
+        @Example(
+            title = "Call an AI agent as a tool",
+            full = true,
+            code = {
+                """
+                    id: ai-agent-with-agent-tools
+                    namespace: company.ai
+
+                    inputs:
+                      - id: prompt
+                        type: STRING
+                        defaults: |
+                          Each flow can produce outputs that can be consumed by other flows. This is a list property, so that your flow can produce as many outputs as you need.
+                          Each output needs to have an ID (the name of the output), a type (the same types you know from inputs, e.g., STRING, URI, or JSON), and a value, which is the actual output value that will be stored in internal storage and passed to other flows when needed.
+                    tasks:
+                      - id: ai-agent
+                        type: io.kestra.plugin.ai.agent.AIAgent
+                        provider:
+                          type: io.kestra.plugin.ai.provider.GoogleGemini
+                          modelName: gemini-2.5-flash
+                          apiKey: "{{ kv('GEMINI_API_KEY') }}"
+                        systemMessage: Summarize the user message, then translate it into French using the provided tool.
+                        prompt: "{{inputs.prompt}}"
+                        tools:
+                          - type: io.kestra.plugin.ai.tool.AIAgent
+                            description: Translation expert
+                            systemMessage: You are an expert in translating text between multiple languages
+                            provider:
+                              type: io.kestra.plugin.ai.provider.GoogleGemini
+                              modelName: gemini-2.5-flash-lite
+                              apiKey: "{{ kv('GEMINI_API_KEY') }}\""""
+            }
+        ),
+    }
+)
+@JsonDeserialize
+@Schema(
+    title = "Call an AI Agent as a tool",
+    description = """
+        This tool allows an LLM to call an AI Agent.
+        Make sure to specify a name and a description so the LLM can understand what it does to decide if it needs to call it."""
+)
+public class AIAgent extends ToolProvider {
+    private static final String TOOL_DESCRIPTION = "This tool allows to call an AI agent named '%s'.";
+
+    @Schema(
+        title = "Agent name",
+        description = "It must be set to a different value than the default in case you want to have multiple agents used as tools in the same task."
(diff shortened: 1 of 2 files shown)


src/main/java/io/kestra/plugin/ai/tool/Agent.java
```diff
@@ -0,0 +1,215 @@
...
+                      - id: ai-agent
+                        type: io.kestra.plugin.ai.agent.AIAgent
+                        provider:
+                          type: io.kestra.plugin.ai.provider.GoogleGemini
+                          modelName: gemini-2.5-flash
+                          apiKey: "{{ kv('GEMINI_API_KEY') }}"
+                        systemMessage: You are an AI agent, summarize the user message then translate it in french using the provided tool.
+                        prompt: "{{inputs.prompt}}"
+                        tools:
+                          - type: io.kestra.plugin.ai.tool.Agent
```
fdelbrayelle: Won't it be confused with `io.kestra.plugin.ai.agent.AIAgent` :-/ ?

fdelbrayelle:
LGTM but I feel like it brings a lot of complexity and not that trivial for the user. What would be nominal use cases for it?

fdelbrayelle (approved):

src/main/java/io/kestra/plugin/ai/tool/Agent.java
loicmathieu: It's a tool, so it would not be inside the same autocompletion result, but yes and no as both are agents, one as a task the other as a tool to be used by another agent.

loicmathieu:
Yes, it's an advanced tool for complex agent use cases (kind of agent workflow where a main agent decide or not to call sub agents).
It's a way to create multi-agents workflow that discuss together

loicmathieu:
Anyway, I'll wait for @anna-geller's feedback before merging to be sure we have the right DSL here

src/main/java/io/kestra/plugin/ai/tool/Agent.java
anna-geller: FWIW, we could align the naming with both to make it distinguishable only via tool/agent subpath, WDYT?
- `io.kestra.plugin.ai.agent.AIAgent`
- `io.kestra.plugin.ai.tool.AIAgent`

src/main/java/io/kestra/plugin/ai/tool/Agent.java
```diff
@@ -0,0 +1,215 @@
...
+
+import static io.kestra.core.utils.Rethrow.throwFunction;
+
+@Getter
+@SuperBuilder
+@NoArgsConstructor
+@Plugin(
+    examples =  {
+        @Example(
+            title = "Call a AI agent as a tool",
```
anna-geller:
```suggestion
            title = "Call an AI agent as a tool",
```

src/main/java/io/kestra/plugin/ai/tool/Agent.java
```diff
@@ -0,0 +1,215 @@
...
+                """
+                    id: ai-agent-with-agent-tools
+                    namespace: company.ai
+
+                    inputs:
+                      - id: prompt
+                        type: STRING
+                        defaults: |
+                          Each flow can produce outputs that can be consumed by other flows. This is a list property, so that your flow can produce as many outputs as you need.
+                          Each output needs to have an id (the name of the output), a type (the same types you know from inputs e.g. STRING, URI or JSON) and value which is the actual output value that will be stored in internal storage and passed to other flows when needed.
```
anna-geller:
```suggestion
                          Each flow can produce outputs that can be consumed by other flows. This is a list property, so that your flow can produce as many outputs as you need.
                          Each output needs to have an ID (the name of the output), a type (the same types you know from inputs, e.g., STRING, URI, or JSON), and a value, which is the actual output value that will be stored in internal storage and passed to other flows when needed.
```

src/main/java/io/kestra/plugin/ai/tool/Agent.java
```diff
@@ -0,0 +1,215 @@
...
+                    tasks:
+                      - id: ai-agent
+                        type: io.kestra.plugin.ai.agent.AIAgent
+                        provider:
+                          type: io.kestra.plugin.ai.provider.GoogleGemini
+                          modelName: gemini-2.5-flash
+                          apiKey: "{{ kv('GEMINI_API_KEY') }}"
+                        systemMessage: You are an AI agent, summarize the user message then translate it in french using the provided tool.
```
anna-geller:
```suggestion
                        systemMessage: Summarize the user message, then translate it into French using the provided tool.
```

src/main/java/io/kestra/plugin/ai/tool/Agent.java
```diff
@@ -0,0 +1,215 @@
...
+                        prompt: "{{inputs.prompt}}"
+                        tools:
+                          - type: io.kestra.plugin.ai.tool.Agent
+                            description: An AI agent expert on translation
```
anna-geller:
```suggestion
                            description: Translation expert
```
what role does description play here? Is it meant for tool discovery?

src/main/java/io/kestra/plugin/ai/tool/Agent.java
```diff
@@ -0,0 +1,215 @@
...
+                            systemMessage: You are an expert system on translation
```
anna-geller:
```suggestion
                            systemMessage: You are an expert in translating text between multiple languages 
```

src/main/java/io/kestra/plugin/ai/tool/Agent.java
```diff
@@ -0,0 +1,215 @@
...
+                              type: io.kestra.plugin.ai.provider.GoogleGemini
+                              modelName: gemini-2.5-flash-lite
+                              apiKey: "{{ kv('GEMINI_API_KEY') }}\""""
+            }
+        ),
+    }
+)
+@JsonDeserialize
+@Schema(
+    title = "Call a AI Agent as a tool",
```
anna-geller:
```suggestion
    title = "Call an AI Agent as a tool",
```

src/main/java/io/kestra/plugin/ai/tool/Agent.java
```diff
@@ -0,0 +1,215 @@
...
+    description = """
+        This tool allows an LLM to call a AI Agent.
+        The agent must have a name and a description so the LLM can understand what it does to decide if it needs to call it."""
```
anna-geller:
```suggestion
        This tool allows an LLM to call an AI Agent.
        Make sure to specify a name and a description so the LLM can understand what it does to decide if it needs to call it."""
```

src/main/java/io/kestra/plugin/ai/tool/AIAgent.java
```diff
@@ -0,0 +1,215 @@
...
+    title = "Call a AI Agent as a tool",
+    description = """
+        This tool allows an LLM to call a AI Agent.
+        The agent must have a name and a description so the LLM can understand what it does to decide if it needs to call it."""
+)
+public class Agent extends ToolProvider {
+    private static final String TOOL_DESCRIPTION = "This tool allows to call an AI agent named '%s'.";
+
+    @Schema(
+        title = "Agent name",
```
anna-geller, line 90:
```suggestion
        name = "Agent name",
```
if it's a `name`, should we call the property a name? You seem to refer it everywhere as name, so maybe worth renaming?

src/test/java/io/kestra/plugin/ai/tool/AgentTest.java
```diff
@@ -0,0 +1,73 @@
...
+                        .build()
+                    )
+                    .build()
+            ))
+            .messages(Property.ofValue(
+                List.of(
+                    ChatCompletion.ChatMessage.builder().type(ChatCompletion.ChatMessageType.SYSTEM).content("You are an AI agent, summarize the user message then translate it in french using the provided tool.").build(),
+                    ChatCompletion.ChatMessage.builder().type(ChatCompletion.ChatMessageType.USER).content("""
+                        Each flow can produce outputs that can be consumed by other flows. This is a list property, so that your flow can produce as many outputs as you need.
+                        Each output needs to have an id (the name of the output), a type (the same types you know from inputs e.g. STRING, URI or JSON) and value which is the actual output value that will be stored in internal storage and passed to other flows when needed.""").build()
```
anna-geller:
```suggestion
                        Each flow can produce outputs that can be consumed by other flows. This is a list property, so that your flow can produce as many outputs as you need.
                        Each output needs to have an `id` (the name of the output), a `type` (the same types you know from inputs e.g., STRING, URI or JSON), and `value`, which is the actual output value that will be stored in internal storage and passed to other flows when needed.""").build()
```

anna-geller (approved):
just minor comments/suggestions added

Thanks for adding this so quickly, looking forward to trying it!

src/main/java/io/kestra/plugin/ai/tool/Agent.java
loicmathieu: Agreed, they provide the same features but in different use site so best to name them the same.

src/main/java/io/kestra/plugin/ai/tool/AIAgent.java
loicmathieu: Here it's the title of the OpenAPI schema not the name of the property which is `name`.

loicmathieu:
@anna-geller I applied your suggestions.
I'll rename the tool to `AIAgent` then merge.

PR #185, opened 2025-10-03
fix(azure-openai): use `maxCompletionTokens` instead of `maxTokens` t…

brian-mulier-p:
…o support newer models

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,2 @@
 version=1.1.0-SNAPSHOT
-kestraVersion=1.0.0
+kestraVersion=1.0.4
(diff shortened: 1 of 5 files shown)


Malaydewangan09 (approved):

fdelbrayelle (approved):

gradle.properties
```diff
@@ -1,2 +1,2 @@
 version=1.1.0-SNAPSHOT
-kestraVersion=[1.0,)
\ No newline at end of file
+kestraVersion=[1.0.4,)
```
loicmathieu: Why? Does it depends on a library upgrade only available since 1.0.4?

loicmathieu (changes requested):
We have other providers that uses OpenAI (Deepseek, OpenRouter) should they also upgrade?

Are you sure maxTokens and maxCompletionTokens are the same?

gradle.properties
brian-mulier-p: yes Azure OpenAI implementation from langchain4j only introduced after an issue I created

brian-mulier-p:
> We have other providers that uses OpenAI (Deepseek, OpenRouter) should they also upgrade?

I pushed the fix, forgot a bit but everything should be ok now, still sad that we are not using the OpenAi task with just putting the parameters we want to fix by default (endpoint for eg.)

> Are you sure maxTokens and maxCompletionTokens are the same?

From my reading it's a simple renaming that is working with all models, newer models don't support maxTokens

PR #191, opened 2025-10-07
feat: add system prompts to AI tasks where it makes sense

### LINKED ISSUE #30: Add system prompts to AI tasks where it makes sense
### Feature description

Some tasks may be better with system messages:

- [x] `Classification` instead of concatenating the user prompt and the classification prompt
- [x] `JSONStructuredExtraction` as an optional additional system prompt

fdelbrayelle:
closes https://github.com/kestra-io/plugin-ai/issues/30

---

---

### Setup Instructions

---

### Contributor Checklist ✅

- [ ] PR Title and commits follows [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [ ] Add a `closes #ISSUE_ID` or `fixes #ISSUE_ID` in the description if the PR relates to an opened issue.
- [ ] Documentation updated (plugin docs from `@Schema` for properties and outputs, `@Plugin` with examples, `README.md` file with basic knowledge and specifics).
- [ ] Setup instructions included if needed (API keys, accounts, etc.).
- [ ] Prefix all rendered properties by `r` not `rendered` (eg: `rHost`).
- [ ] Use `runContext.logger()` to log enough important infos where it's needed and with the best level (DEBUG, INFO, WARN or ERROR).

⚙️ **Properties**
- [ ] Properties are declared with `Property<T>` carrier type, do **not** use `@PluginProperty`.
- [ ] Mandatory properties must be annotated with `@NotNull` and checked during the rendering.
- [ ] You can model a JSON thanks to a simple `Property<Map<String, Object>>`.

🌐 **HTTP**
- [ ] Must use Kestra’s internal HTTP client from `io.kestra.core.http.client`

📦 **JSON**
- [ ] If you are serializing response from an external API, you may have to add a `@JsonIgnoreProperties(ignoreUnknown = true)` at the mapped class level. So that we will avoid to crash the plugin if the provider add a new field suddenly.
- [ ] Must use Jackson mappers provided by core (`io.kestra.core.serializers`)

✨ **New plugins / subplugins**
- [ ] Make sure your new plugin is configured like mentioned [here](https://kestra.io/docs/plugin-developer-guide/gradle#mandatory-configuration).
- [ ] Add a `package-info.java` under each sub package respecting [this format](https://github.com/kestra-io/plugin-odoo/blob/main/src/main/java/io/kestra/plugin/odoo/package-info.java) and choosing the right category.
- [ ] Icons added in `src/main/resources/icons` in SVG format and not in thumbnail (keep it big):
  - `plugin-icon.svg`
  - One icon per package, e.g. `io.kestra.plugin.aws.svg`
  - For subpackages, e.g. `io.kestra.plugin.aws.s3`, add `io.kestra.plugin.aws.s3.svg`
    See example [here](https://github.com/kestra-io/plugin-elasticsearch/blob/master/src/main/java/io/kestra/plugin/elasticsearch/Search.java#L76).
- [ ] Use `"{{ secret('YOUR_SECRET') }}"` in the examples for sensible infos such as an API KEY.
- [ ] If you are fetching data (one, many or too many), you must add a `Property<FetchType> fetchType` to be able to use `FETCH_ONE`, `FETCH` and even `STORE` to store big amount of data in the internal storage.
- [ ] Align the `"""` to close examples blocks with the flow id.

🧪 **Tests**
- [ ] Unit Tests added or updated to cover the change (using the `RunContext` to actually run tasks).
- [ ] Add sanity checks if possible with a YAML flow inside `src/test/resources/flows`.
- [ ] Avoid disabling tests for CI. Instead, configure a local environment whenever it's possible with `.github/setup-unit.sh` (which can be executed locally and in the CI) all along with a new `docker-compose-ci.yml` file (do **not** edit the existing `docker-compose.yml`).
- [ ] Provide screenshots from your QA / tests locally in the PR description. The goal here is to use the JAR of the plugin and directly test it locally in Kestra UI to ensure it integrates well.

📤 **Outputs**
- [ ] Do not send back as outputs the same infos you already have in your properties.
- [ ] If you do not have any output use `VoidOutput`.
- [ ] Do not output twice the same infos (eg: a status code, an error code saying the same thing...).

### DIFF
--- a/src/main/java/io/kestra/plugin/ai/completion/Classification.java
+++ b/src/main/java/io/kestra/plugin/ai/completion/Classification.java
@@ -1,5 +1,7 @@
 package io.kestra.plugin.ai.completion;
 
+import dev.langchain4j.data.message.AiMessage;
+import dev.langchain4j.data.message.SystemMessage;
 import dev.langchain4j.data.message.UserMessage;
 import dev.langchain4j.model.chat.ChatModel;
 import dev.langchain4j.model.chat.response.ChatResponse;
@@ -21,6 +23,7 @@ import lombok.*;
 import lombok.experimental.SuperBuilder;
 import org.slf4j.Logger;
 
+import java.util.ArrayList;
 import java.util.List;
 
 @SuperBuilder
@@ -39,22 +42,26 @@ import java.util.List;
             full = true,
             code = {
                 """
-                id: text_categorization
-                namespace: company.ai
+                    id: text_categorization
+                    namespace: company.ai
 
-                tasks:
-                  - id: categorize
-                    type: io.kestra.plugin.ai.completion.Classification
-                    prompt: "Categorize the sentiment of: I love this product!"
-                    classes:
-                      - positive
-                      - negative
-                      - neutral
-                    provider:
-                      type: io.kestra.plugin.ai.provider.GoogleGemini
-                      apiKey: "{{ kv('GEMINI_API_KEY') }}"
-                      modelName: gemini-2.5-flash
-                """
+                    tasks:
+                      - id: categorize
+                        type: io.kestra.plugin.ai.completion.Classification
+                        messages:
+                          - type: SYSTEM
+                            content: You are a sentiment analysis assistant. Classify text as positive, negative or neutral.
+                          - type: USER
+                            content: "I absolutely love this product, it's fantastic!"
+                        classes:
+                          - positive
+                          - negative
+                          - neutral
+                        provider:
+                          type: io.kestra.plugin.ai.provider.GoogleGemini
+                          apiKey: "{{ kv('GEMINI_API_KEY') }}"
+                          modelName: gemini-2.0-flash
+                    """
             }
         )
     },
@@ -62,9 +69,12 @@ import java.util.List;
 )
 public class Classification extends Task implements RunnableTask<Classification.Output> {
 
-    @Schema(title = "Text prompt", description = "The input text to classify")
+    @Schema(
+        title = "Chat Messages",
+        description = "The list of chat messages for the current conversation. There can be only one system message, and the last message must be a user message."
+    )
     @NotNull
-    private Property<String> prompt;
+    private Property<List<io.kestra.plugin.ai.domain.ChatMessage>> messages;
 
     @Schema(title = "Classification Options", description = "The list of possible classification categories")
     @NotNull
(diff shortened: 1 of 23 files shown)


src/main/java/io/kestra/plugin/ai/completion/Classification.java
```diff
@@ -39,32 +43,39 @@
...
 )
 public class Classification extends Task implements RunnableTask<Classification.Output> {
 
-    @Schema(title = "Text prompt", description = "The input text to classify")
+    @Schema(
+        title = "Chat Messages",
+        description = "The list of chat messages for the current conversation. There can be only one system message, and the last message must be a user message."
+    )
     @NotNull
-    private Property<String> prompt;
```
loicmathieu, line 67: Please keep a single prompt.
The idea was to use a system message under the cover to add the classification instruction instead of concatenating them inside the user prompt.

Another way to do it would be to add the instruction as default value in a `systemMessage` so the user could update the instruction if he want.

src/main/java/io/kestra/plugin/ai/completion/ImageGeneration.java
loicmathieu, line 1: Sorry, I was wrong, this task should not have a system message as it's not possible to pass one via the Langchain4J API.

src/main/java/io/kestra/plugin/ai/completion/JSONStructuredExtraction.java
```diff
@@ -106,9 +117,12 @@
 )
 public class JSONStructuredExtraction extends Task implements RunnableTask<JSONStructuredExtraction.Output> {
 
-    @Schema(title = "Text prompt", description = "The input prompt for the AI model")
+    @Schema(
+        title = "Chat Messages",
+        description = "The list of chat messages for the extraction context. There can be only one system message, and the last message must be a user message."
+    )
     @NotNull
-    private Property<String> prompt;
```
loicmathieu, line 111: Please keep a single prompt.
The idea was to use a system message under the cover to add the structured instruction instead of concatenating them inside the user prompt.

src/main/java/io/kestra/plugin/ai/rag/ChatCompletion.java
loicmathieu, line 1: Why changing it? It already contains a system message so I'll revert all changes.

loicmathieu (changes requested):

src/main/java/io/kestra/plugin/ai/completion/Classification.java
```diff
@@ -66,6 +68,9 @@ public class Classification extends Task implements RunnableTask<Classification.
     @NotNull
     private Property<String> prompt;
 
+    @Schema(title = "Optional system message", description = "Instruction message for the model. Defaults to a standard classification instruction.")
+    private Property<String> systemMessage;
```
loicmathieu: Please add the default instructions as a default property value so it is documented properly

loicmathieu: The classes should be added  to additional rendering variable.
That's more work, that's why I said it would be optional ;)

src/main/java/io/kestra/plugin/ai/completion/JSONStructuredExtraction.java
```diff
@@ -45,71 +46,75 @@
...
 )
 public class JSONStructuredExtraction extends Task implements RunnableTask<JSONStructuredExtraction.Output> {
 
-    @Schema(title = "Text prompt", description = "The input prompt for the AI model")
-    @NotNull
+    @Schema(title = "Text prompt", description = "The input text for structured JSON extraction.")
     private Property<String> prompt;
 
+    @Schema(title = "System message", description = "Optional system instruction for the model.")
+    private Property<String> systemMessage;
```
loicmathieu: same here

loicmathieu (changes requested):

src/main/java/io/kestra/plugin/ai/completion/Classification.java
```diff
@@ -85,28 +97,25 @@ public class Classification extends Task implements RunnableTask<Classification.
     public Classification.Output run(RunContext runContext) throws Exception {
         Logger logger = runContext.logger();
 
-        // Render input properties
-        String renderedPrompt = runContext.render(prompt).as(String.class).orElseThrow();
-        List<String> renderedClasses = runContext.render(classes).asList(String.class);
```
loicmathieu, line 90: You still need to render calsses and use render classes to render the system prompt as we didn't render recursively

src/test/java/io/kestra/plugin/ai/completion/ClassificationTest.java
```diff
@@ -35,13 +35,13 @@ void testClassificationGemini() throws Exception {
...
             "apiKey", GEMINI_API_KEY,
-            "modelName", "gemini-1.5-flash"
-
+            "modelName", "gemini-2.0-flash"
         ));
 
         Classification task = Classification.builder()
             .prompt(Property.ofExpression("{{ prompt }}"))
-            .classes(Property.ofValue(List.of("{{ classes }}")))
+            .systemMessage(Property.ofValue("You are a text classification assistant."))
```
loicmathieu, line 43: I'm surprised the test works without passing the class inside the system message

loicmathieu (approved):
I approved it to avoid a new back and forth but please fix the rendering of classes before merging

PR #201, opened 2025-10-14
fix: missing metrics

### LINKED ISSUE #11988: Missing metrics in AI plugin
### Describe the issue

- [ ] plugin-ai/src/main/java/io/kestra/plugin/ai/AIUtils.java

### Environment

- Kestra Version: develop

Raj-G07:
- Closes https://github.com/kestra-io/kestra/issues/11988

### DIFF
--- a/src/main/java/io/kestra/plugin/ai/AIUtils.java
+++ b/src/main/java/io/kestra/plugin/ai/AIUtils.java
@@ -4,6 +4,8 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import dev.langchain4j.agent.tool.ToolSpecification;
 import dev.langchain4j.service.tool.ToolExecutor;
 import io.kestra.core.exceptions.IllegalVariableEvaluationException;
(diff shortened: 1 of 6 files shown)


src/main/java/io/kestra/plugin/ai/AIUtils.java
```diff
@@ -17,6 +19,29 @@
 
 import static io.kestra.core.utils.Rethrow.throwConsumer;
 
+@Plugin(
```
fdelbrayelle: @Raj-G07 ⚠️ No, you don't have to add a `@Plugin` annotation here since it's just a util class.

Instead you have to search for all calls to `sendMetrics` method:

Image

For those which are actual plugins with an existing `@Plugin` annotation, then you'll have to add the `@Metric` annotations with appropriate names.

Raj-G07: My bad 😐

fdelbrayelle (approved):

PR #204, opened 2025-10-17
feat: allow defining custom topology nodes from tasks

loicmathieu:
Part-of: https://github.com/kestra-io/kestra-ee/issues/4729

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,2 @@
 version=1.1.0-SNAPSHOT
-kestraVersion=1.0.4
+kestraVersion=1.1.0-SNAPSHOT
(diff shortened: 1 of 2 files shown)


loicmathieu:
Decision has been made to make another PoC with plugin-emitted Vue components

fdelbrayelle:
@loicmathieu What is the remaining work finally on this one? Don't we want to add a local `ui` folder here like @elevatebart is doing for plugin-redis ([PR](https://github.com/kestra-io/plugin-redis/pull/141))?

loicmathieu:
This is an old prototype that is not aligned anymore with what we want to do for plugin UI

PR #245, opened 2026-01-22
chore: try 25 CI with Java 21

loicmathieu:
WIP PR, will be deleted


loicmathieu:
This proves that using an action in Java 25 sucessfully builds a plugin in 21.
Closing it now.

PR #261, opened 2026-03-02
Multi Modal Input Support

jrsperry:
### Summary
This PR adds multimodal support to `io.kestra.plugin.ai.completion.ChatCompletion` so a `USER` message can include structured `contents` blocks (`TEXT`, `IMAGE`, `PDF`) in addition to the existing `content` string field.

### Why
Users need to send uploaded documents/images directly in chat prompts (for analysis/extraction), while preserving current text-only behavior for existing flows.

### What changed
- Added `contents` support to `ChatMessage` with content blocks:
  - `type`: `TEXT` (default), `IMAGE`, `PDF`
  - `text`: for text blocks
  - `uri`: for file-backed image/pdf blocks
- Kept backward compatibility:
  - Existing `content` field still works exactly as before.
  - If `contents` is not set, `content` is treated as a single `TEXT` block.
- Updated chat message conversion in `ChatCompletion`:
  - `SYSTEM` and `AI` messages remain text-only.
  - `USER` messages can now contain mixed blocks.
- File handling behavior:
  - `IMAGE` and `PDF` blocks must use uploaded Kestra file URIs (`kestra://...`) from `FILE` inputs.
  - Task loads bytes from Kestra storage and base64-encodes payloads before sending to LangChain4j.
  - Image MIME type is detected from content; PDF uses `application/pdf`.
- Updated assistant call path to send `List<Content>` (multimodal) instead of `singleText()`.

### Version requirement
`1.3.0-SNAPSHOT` is required, because it brings LangChain4j `1.11`.  
LangChain4j `1.10` has known issues for PDF inputs when using `AiServices`.

### Validation
- Added integration tests for multimodal conversion and validation behavior:
  - Legacy text-only input remains default behavior.
  - Mixed `TEXT + IMAGE + PDF` conversion from uploaded files works.
  - Invalid non-`kestra://` file URI is rejected.
  - Non-text blocks on `SYSTEM` messages are rejected.
- Manual provider validation performed with `pdf`, `jpeg`, and `png`.

### Sample flow (PDF)
```yaml
id: test_analyze_pdf
namespace: company.ai

inputs:
  - id: pdf
    type: FILE
  - id: question
    type: STRING
    defaults: "Summarize this PDF and list the top 5 key points."

tasks:
  - id: analyze
    type: io.kestra.plugin.ai.completion.ChatCompletion
    provider:
      type: io.kestra.plugin.ai.provider.OpenAI
      apiKey: "{{ secret('OPENAI_API_KEY') }}"
      modelName: gpt-5-mini
    messages:
      - type: SYSTEM
        content: "You are a precise document analyst. Be concise and factual."
      - type: USER
        contents:
          - text: "{{ inputs.question }}"
          - type: PDF
            uri: "{{ inputs.pdf }}"
```

### Contributor Checklist ✅
- [ ] I have read and followed the [plugin contribution guidelines](https://kestra.io/docs/plugin-developer-guide/contribution-guidelines)


jrsperry:
So I only did this multi modal support for ChatCompletion, looking for feedback, happy to add it for Agent, Classification, JSONStructuredExtraction as well and pull some of the utils out into a helper class. Looking for feedback about the structure of the content blocks. 

I have this running actively (only needed the lower level ChatCompletion)

gradle.properties
```diff
@@ -1,2 +1,2 @@
 version=1.7.1-SNAPSHOT
-kestraVersion=1.2.3
\ No newline at end of file
+kestraVersion=1.3.0-SNAPSHOT
```
fdelbrayelle: Hi @jrsperry :wave: What do you need from 1.3.0 again?

```suggestion
kestraVersion=1.3.0
```

jrsperry: langchain4j 1.10 has issues with multi modal inputs when using AiServices. It blows out the token limit and misrepresents them. The underlying chat models works fine, but we use AiServices in the plugins, thankfully this problem is solved in 1.11 of langchain4j. 

I discovered this bug and confirmed with a test that AiServices was manipulating those messages as the token input with a pdf was like 15 times higher (and AI couldn't understand either) with AiServices vs ChatModel. Again, that's solved in 1.11 thankfully.

src/main/java/io/kestra/plugin/ai/domain/ChatMessage.java
```diff
@@ -2,6 +2,45 @@
...
+import java.util.List;
+
 @Builder
-public record ChatMessage(ChatMessageType type, String content) {
-}
\ No newline at end of file
+public record ChatMessage(
+    ChatMessageType type,
+    String content,
+    List<ContentBlock> contents
```
loicmathieu:
```suggestion
    List<ContentBlock> contentBlocks
```

src/main/java/io/kestra/plugin/ai/completion/ChatCompletion.java
loicmathieu: Please describe that a ChatMessage can either be a text (using content) or a multi-block multimodal message

src/main/java/io/kestra/plugin/ai/domain/ChatMessage.java
```diff
@@ -2,6 +2,45 @@
 
 import lombok.Builder;
 
+import java.util.List;
+
 @Builder
-public record ChatMessage(ChatMessageType type, String content) {
-}
\ No newline at end of file
+public record ChatMessage(
```
loicmathieu, line 13: I wonder if we should not add validation so either content or contentBlocks are filled.
We may also add documentation here as it would not be self-explaning anymore

loicmathieu (changes requested):
Some realy feedback, the final review would be from @fdelbrayelle

fdelbrayelle:
@jrsperry Is it ready to review? I saw you put "DRAFT" in the PR's title.

jrsperry:
@fdelbrayelle  I was waiting to get feedback  (particularly around the message content definitions) before applying this to some of the tasks (such as agent, json structured extraction, etc).

fdelbrayelle:
@jrsperry It sounds good to me so far :) So you can extend it to the other tasks you told.

jrsperry:
@fdelbrayelle  I believe it's ready for review. I decided against doing the work for the AiAgent as I wasn't looking to tackle the memory storage problem.

src/main/java/io/kestra/plugin/ai/domain/ChatMessage.java
fdelbrayelle: @aj-emerich Could you check this one please :pray: ?

src/main/java/io/kestra/plugin/ai/completion/ChatCompletion.java
```diff
@@ -105,6 +106,30 @@
...
+                    provider:
+                      type: io.kestra.plugin.ai.provider.OpenAI
+                      apiKey: "{{ secret('OPENAI_API_KEY') }}"
+                      modelName: gpt-4o-mini
+                    messages:
+                      - type: USER
+                        contentBlocks:
+                          - text: Summarize this document.
+                          - type: PDF
+                            uri: "{{ outputs.upload.uri }}"
```
aj-emerich, line 143: Does this make sense? Where does the upload come from?

src/main/java/io/kestra/plugin/ai/domain/ChatMessage.java
```diff
@@ -1,7 +1,62 @@
...
-}
\ No newline at end of file
+@Schema(
+    title = "Chat Message",
+    description = "A chat message payload. Use either `content` for plain text or `contentBlocks` for multimodal content blocks."
+)
+public record ChatMessage(
+    @Schema(title = "Message type")
+    ChatMessageType type,
+    @Schema(title = "Text content", description = "Plain text message content. Must be set only when `contentBlocks` is not set.")
```
aj-emerich: @jrsperry I think it makes more sense to communicate that `content` and `contentBlocks` are "mutually exclusive" with each other in their descriptions.

aj-emerich (changes requested):

src/main/java/io/kestra/plugin/ai/domain/ChatMessage.java
```diff
@@ -1,7 +1,62 @@
...
+                    .text(content)
+                    .build()
+            );
+        }
+
+        return List.of();
+    }
+
+    @Builder
+    public record ContentBlock(
```
fdelbrayelle, line 47: @aj-emerich Shouldn't we add also docs on each attribute here?

src/main/java/io/kestra/plugin/ai/completion/ChatCompletion.java
loicmathieu: It would be better to use a smart URI here so we allow kestra, local and namespace files.
For Kestra files, this would comes from a previous task that download the file for ex?

Another good imporvement (maybe for later) would be to allow setting it as either an URI or a base64 string

src/main/java/io/kestra/plugin/ai/completion/ChatCompletion.java
```diff
@@ -273,17 +297,50 @@ public ChatCompletion.Output run(RunContext runContext) throws Exception {
...
-            .map(dto -> switch (dto.type()) {
-                case SYSTEM -> SystemMessage.systemMessage(dto.content());
-                case AI ->  AiMessage.aiMessage(dto.content());
-                case USER ->  UserMessage.userMessage(dto.content());
-            })
-            .toList();
+    private List<ChatMessage.ContentBlock> resolveContents(ChatMessage message) {
+        List<ChatMessage.ContentBlock> contents = message.effectiveContents();
+        if (contents.isEmpty()) {
+            throw new IllegalArgumentException("Message type " + message.type() + " must define either `content` (legacy) or `contentBlocks`.");
```
loicmathieu:
```suggestion
            throw new IllegalArgumentException("Message type " + message.type() + " must define either `content` or `contentBlocks`.");
```

It's not legacy as it's not deprecated

src/main/java/io/kestra/plugin/ai/completion/Classification.java
```diff
@@ -89,10 +90,16 @@
...
-    @NotNull
+    @Schema(title = "Text prompt", description = "Text input to classify. Use either `prompt` or `promptContentBlocks`.")
     private Property<String> prompt;
 
+    @Schema(
+        title = "Prompt content blocks",
+        description = "Multimodal input blocks for classification (TEXT, IMAGE, PDF). Use either `prompt` or `promptContentBlocks`."
+    )
+    @Nullable
+    private Property<List<ChatMessage.ContentBlock>> promptContentBlocks;
```
loicmathieu:
```suggestion
    private Property<List<ChatMessage.ContentBlock>> contentBlocks;
```

It seems more straighforward

src/main/java/io/kestra/plugin/ai/domain/ChatMessage.java
aj-emerich: Yes, every property/attribute should have a straightforward `title` and a `description` that gives an example or a default that gives more context to its use.

src/main/java/io/kestra/plugin/ai/domain/ChatMessage.java
```diff
@@ -1,7 +1,62 @@
...
+    ChatMessageType type,
+    @Schema(title = "Text content", description = "Plain text message content. Must be set only when `contentBlocks` is not set.")
+    String content,
+    @Schema(title = "Content blocks", description = "Multimodal message blocks (TEXT, IMAGE, PDF). Must be set only when `content` is not set.")
+    List<ContentBlock> contentBlocks
+) {
+    public ChatMessage {
+        boolean hasTextContent = content != null && !content.isBlank();
+        boolean hasBlockContent = contentBlocks != null && !contentBlocks.isEmpty();
+        if (hasTextContent == hasBlockContent) {
```
fdelbrayelle, line 24: ⚠️ The compact constructor enforces that `hasTextContent == hasBlockContent` must be false, meaning exactly one must be set. But when Kestra deserializes YAML like `content: "hello"`, the
contentBlocks field will be null and so content will be non-blank: so `true == false` will be
valid.

However, the issue is the equality check:
- if both are null/empty (someone
constructs `ChatMessage.builder().type(USER).build()`), then `false == false` is `true` which
correctly throws
- but now if someone provides content: "" (blank) and no blocks, then `false
== false` throws. This edge case is fine but...

The real concern is a breaking change to serialization. The old `ChatMessage` had only type and content. Any existing persisted flows or serialized messages with just content will still work, but the record now has a 3 arguments constructor. Could you check that Kestra's Jackson deserialization handles this gracefully through a unit test (the new contentBlocks field defaulting to null when absent from YAML/JSON)?

src/main/java/io/kestra/plugin/ai/completion/CompletionInputContentUtils.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private static URI parseUri(String raw, String blockType) {
+        try {
+            return URI.create(raw);
+        } catch (IllegalArgumentException e) {
+            throw new IllegalArgumentException("Invalid " + blockType + " uri: `" + raw + "`", e);
+        }
+    }
+
+    private static byte[] resolveKestraFileBytes(RunContext runContext, URI uri, String blockType) throws Exception {
```
loicmathieu: Here, you can use `new URIFetcher(uri).fetch(runContext)` for smart URI.
Smart URI can be either a Kestra URI, a local file path or a namespace file.
Other URIs may be supported transparently later.

Such URI should be documented with the supported scheme, i'm not sure how we usually do it, maybe @fdelbrayelle or @aj-emerich know

src/main/java/io/kestra/plugin/ai/completion/JSONStructuredExtraction.java
```diff
@@ -133,9 +134,16 @@
...
-    @Schema(title = "Text prompt", description = "The input text for structured JSON extraction.")
+    @Schema(title = "Text prompt", description = "Text input for structured JSON extraction. Use either `prompt` or `promptContentBlocks`.")
     private Property<String> prompt;
 
+    @Schema(
+        title = "Prompt content blocks",
+        description = "Multimodal input blocks for extraction (TEXT, IMAGE, PDF). Use either `prompt` or `promptContentBlocks`."
+    )
+    @Nullable
+    private Property<List<ChatMessage.ContentBlock>> promptContentBlocks;
```
loicmathieu:
```suggestion
    private Property<List<ChatMessage.ContentBlock>> contentBlocks;
```

For simplicity

src/main/java/io/kestra/plugin/ai/completion/Classification.java
```diff
@@ -89,10 +90,16 @@
 )
 public class Classification extends Task implements RunnableTask<Classification.Output> {
 
-    @Schema(title = "Text prompt", description = "The input text to classify")
-    @NotNull
```
fdelbrayelle, line 98: To keep consistency with `promptContentBlocks`, could you add a `@Nullable` here?

loicmathieu (approved):
I added a bunch of suggestion for improvements but I'll approve it as globally it makes sense to me: good addition.
However it misses tests (the existing tests for all models should be updated to include one with the new contentBlocks property).

I'll let @fdelbrayelle do the final validation

src/main/java/io/kestra/plugin/ai/completion/JSONStructuredExtraction.java
```diff
@@ -133,9 +134,16 @@
 )
 public class JSONStructuredExtraction extends Task implements RunnableTask<JSONStructuredExtraction.Output> {
 
-    @Schema(title = "Text prompt", description = "The input text for structured JSON extraction.")
+    @Schema(title = "Text prompt", description = "Text input for structured JSON extraction. Use either `prompt` or `promptContentBlocks`.")
     private Property<String> prompt;
```
fdelbrayelle, line 153: To keep consistency with `promptContentBlocks`, could you add a `@Nullable` here?

src/main/java/io/kestra/plugin/ai/completion/JSONStructuredExtraction.java
```diff
@@ -230,4 +239,15 @@ public static JsonObjectSchema buildDynamicSchema(List<String> fields) {
         schemaBuilder.required(fields);
         return schemaBuilder.build();
     }
+
+    private void validatePromptInput(String prompt, List<ChatMessage.ContentBlock> promptContentBlocks) {
```
fdelbrayelle: Move it to `CompletionInputContentUtils` since it's shared between this class and `JSONStructuredExtraction`.

src/main/java/io/kestra/plugin/ai/completion/ChatCompletion.java
aj-emerich: Yeah, there needs to be a file source task if using outputs, or switch to local or namespace file URI otherwise.

src/main/java/io/kestra/plugin/ai/completion/CompletionInputContentUtils.java
```diff
@@ -0,0 +1,114 @@
...
+            return inputStream.readAllBytes();
+        } catch (Exception e) {
+            throw new IllegalArgumentException("Unable to read " + blockType + " file from Kestra storage uri `" + uri + "`.", e);
+        }
+    }
+
+    private static String resolveImageMediaType(byte[] bytes) throws Exception {
+        String mediaType;
+        try (ByteArrayInputStream stream = new ByteArrayInputStream(bytes)) {
+            mediaType = URLConnection.guessContentTypeFromStream(stream);
```
fdelbrayelle: Not sure `URLConnection.guessContentTypeFromStream` properly recognize every image type, especially JPEG and WebP. Could you check it in a unit test for these types but also for PNG and GIF?

OK maybe consider to add directly a mimeType attribute in `ContentBlock`?

fdelbrayelle (commented):
⚠️ Some important parts to change.

Plus, could you add tests for Classification and JSONStructuredExtension multimodal paths?

src/main/java/io/kestra/plugin/ai/domain/ChatMessage.java
fdelbrayelle: @jrsperry Could you handle it and then @aj-emerich will be able to review it. Thanks!

src/main/java/io/kestra/plugin/ai/domain/ChatMessage.java
loicmathieu: Deserialization will put null on the added attribute.
Reordering attribute would break everything.

fdelbrayelle:
Hello @jrsperry :wave: Did you have any time to take into account our feedback? Thanks :pray:

jrsperry:
just getting back to this, was OOO. looks fine to me.

fdelbrayelle:
OK @jrsperry tell us once you'll have taken our feedbacks :)

jrsperry:
@fdelbrayelle 

I think I may have addressed most concerns.

fdelbrayelle (approved):

PR #323, opened 2026-05-26
feat: Add DockerModel provider for Docker Model Runner

### LINKED ISSUE #322: feat: Add DockerModel provider for Docker Model Runner
## Overview

Add a `DockerModel` provider to `plugin-ai` that routes inference requests to a locally running **Docker Model Runner** instance.

Docker Model Runner ([docs](https://docs.docker.com/ai/model-runner/)) exposes an **OpenAI-compatible REST API** at `http://localhost:12434/engines/v1`, requiring no authentication. This means `DockerModel` can extend the existing `OpenAICompliantProvider` base class and re-use the full provider infrastructure with minimal new code.

---

## Motivation

Docker Model Runner is now a first-class Docker feature available out-of-the-box on Docker Desktop and Docker Engine (Linux). For Kestra users who run local or on-prem workflows, it provides a zero-friction path to self-hosted LLM inference without managing a separate Ollama or LocalAI stack. Adding a dedicated provider makes the integration explicit and discoverable.

---

## Link with plugin-docker

`plugin-docker` ([kestra-io/plugin-docker#119](https://github.com/kestra-io/plugin-docker/issues/119)) adds dedicated tasks (`docker.model.Pull`, `docker.model.List`, etc.) to manage the Docker Model Runner model lifecycle from a Kestra flow.

The separation of concerns is intentional: **plugin-docker owns model lifecycle** (pull, configure, delete), **plugin-ai owns inference**. A typical flow chains both:

```yaml
tasks:
  - id: pull_model
    type: io.kestra.plugin.docker.model.Pull
    model: ai/smollm2

  - id: ask
    type: io.kestra.plugin.ai.completion.ChatCompletion
    provider:
      type: io.kestra.plugin.ai.provider.DockerModel
      modelName: ai/smollm2
    messages:
      - role: user
        content: What is Docker Model Runner?
```

---

### New class: `io.kestra.plugin.ai.provider.DockerModel`

Extends `OpenAICompliantProvider` (same parent as `OpenAI`, `AzureOpenAI`).

**Properties:**
| Property | Type | Default | Notes |
|---|---|---|---|
| `modelName` | `Property<String>` | required | e.g. `ai/smollm2`, `ai/llama3.2:3b-instruct-q4_k_m`, `hf.co/bartowski/Llama-3.2-1B-Instruct-GGUF` |
| `baseUrl` | `Property<String>` | `http://localhost:12434/engines/v1` | Override for Docker Engine remote (e.g. `http://172.17.0.1:12434/engines/v1`) |
| `apiKey` | `Property<String>` | `"not-needed"` (default) | Accepted but ignored by DMR; required by OpenAI SDK |

**Supported capabilities:**
- `chatModel()` — chat completions (`/engines/v1/chat/completions`)
- `embeddingModel()` — text embeddings (`/engines/v1/embeddings`)
- `imageModel()` — image generation via Diffusers (`/engines/diffusers/v1/images/generations`, requires a `diffuser`-tagged model such as `ai/stable-diffusion`)

**Behaviour notes:**
- No API key is required by Docker Model Runner; the `apiKey` field should default to `"not-needed"` and be marked optional in the schema
- Base URL for containers should be `http://model-runner.docker.internal/engines/v1` (Docker Desktop) or `http://172.17.0.1:12434/engines/v1` (Docker Engine)
- The `baseUrl` field should be prominently documented with both variants

---

## Acceptance Criteria

- [ ] `DockerModel` provider implemented extending `OpenAICompliantProvider`
- [ ] `baseUrl` defaults to `http://localhost:12434/engines/v1`; clearly documented with container-internal variants
- [ ] `apiKey` is optional, defaults to `"not-needed"`
- [ ] `chatModel()`, `embeddingModel()`, and `imageModel()` implemented (image generation routes to `/engines/diffusers/v1/`)
- [ ] `@Schema`, `@Plugin`, and `@Example` annotations with at least one working flow example
- [ ] Unit tests covering property rendering, default URL, and optional API key
- [ ] Integration tests annotated `@DockerModelRunnerTest` (skippable when DMR is unavailable)

---

## References

- [Docker Model Runner docs](https://docs.docker.com/ai/model-runner/)
- OpenAI-compatible base URL: `http://localhost:12434/engines/v1`
- Anthropic-compatible base URL: `http://localhost:12434/anthropic`
- Ollama-compatible base URL: `http://localhost:12434`
- Supported model list: `ai/*` namespace on Docker Hub (74+ repos)
- Existing similar provider for reference: `io.kestra.plugin.ai.provider.Ollama`, `io.kestra.plugin.ai.provider.LocalAI`
- Companion issue: [kestra-io/plugin-docker#119](https://github.com/kestra-io/plugin-docker/issues/119) — model lifecycle management tasks

rkampani:
Closes #322


fdelbrayelle:
Could you do a full QA with what I've done on the other PR too before we merge this one please @rkampani ? https://github.com/kestra-io/plugin-docker/pull/123 Thanks :pray:

rkampani:
@fdelbrayelle : https://github.com/kestra-io/plugin-ai/pull/323#issuecomment-4829787420 It looks compatible with other PR.

fdelbrayelle:
@rkampani Could you run a full test with a full flow with the plugin-docker PR please?

MilosPaunovic:
Hey @rkampani, are there any updates on this?

rkampani:
> Hey @rkampani, are there any updates on this?

-  Full live smoke test through a Kestra instance (1.3.29, Docker): built this branch's shadow jar, loaded it into Kestra alongside plugin-docker#123's jar, confirmed both io.kestra.plugin.ai.provider.DockerModel and
-  Ran an actual flow end-to-end (ChatCompletion + DockerModel provider) via the Kestra API — execution SUCCESS, with baseUrl/modelName correctly resolved from RunContext, a real HTTP round trip made from inside the Kestra container and the response correctly parsed into ChatCompletion.Output (text, token usage, finish reason).

Not Validated : 

- If Need DMR validation before merge , i would requires someone with an M-series Mac or Model Runner CI to run the integration test.

jymaire:
Thanks for this — it's a well-scoped contribution that fits the existing provider architecture nicely, and the `@Schema`/`@Plugin`/`@Example` annotations are thorough. A few things to address before we can merge.

First, a note on CI: the red ❌ is **not** your fault. The Java tests ran 184 / 107 passed / 77 skipped / **0 failed** — the job fails on the check-run reporting step (`Resource not accessible by integration`, a token permission issue). No need to chase it.

---

### 1. `apiKey` field shadowing leaves the parent's `@NotNull` field permanently null

`DockerModel` re-declares `private Property<String> apiKey`, but `OpenAICompliantProvider` already declares `@NotNull private Property<String> apiKey`. Java fields don't override — they shadow. `@SuperBuilder` and Jackson both bind to the most-derived field, so the parent's slot is never populated:

```
child apiKey field  = set-by-builder
parent apiKey field = null
getApiKey()         = set-by-builder
validator → 1 violation: apiKey: must not be null
```

To be clear about the blast radius: **this does not break flows today.** `ChatCompletion.provider` has no `@Valid`, so validation doesn't cascade into the provider (validating a full task with a `DockerModel` yields 0 violations), and the generated JSON schema is clean — `required: ["modelName", "type"]`, no duplicate properties. But it leaves a dead `@NotNull` and a permanently-null inherited field, which is a trap for whoever later adds `@Valid` or reads `super.apiKey`.

The cleanest fix also removes the need for the new `resolveApiKey` hook entirely. Note that the identical `baseUrl` shadowing in `DeepSeek` works fine — because the parent reads it through the **getter** `getBaseUrl()`. `apiKey` breaks only because the parent reads the **field** `this.apiKey`. So:

- change `OpenAICompliantProvider` to call `getApiKey()` instead of `this.apiKey`,
- drop the `protected resolveApiKey(...)` hook (it's new API surface on a shared base class added for a single subclass),
- let `DockerModel` just declare its defaulted field exactly like `DeepSeek` does,
- and drop the now-dead `@NotNull` (or move requiredness into an `orElseThrow` that names the property).

### 2. `imageModel()` silently misroutes on custom base URLs

```java
String diffuserUrl = resolvedBaseUrl.replace(CHAT_PATH, DIFFUSER_PATH);
```

For any `baseUrl` that doesn't contain the literal `/engines/v1` — a reverse proxy, an enterprise gateway like `https://gateway/dmr` — `.replace()` is a no-op and image requests are sent to the chat endpoint, surfacing later as an opaque error from the remote. Please fail fast instead, with a message naming the property and the offending value.

### 3. The two `imageModel` tests don't test `imageModel()`

Neither `imageModel_shouldRouteToDefaultDiffuserEndpoint` nor `imageModel_withCustomBaseUrl_shouldStillRewriteToDiffuserPath` ever calls `provider.imageModel(runContext)`. Both assert that `"...".replace(a, b)` returns the expected string — i.e. they test `java.lang.String`. The first builds a `provider` and then never uses it. **Both pass if `imageModel()` is deleted outright.**

WireMock is already wired up in this class, so please stub `/engines/diffusers/v1/images/generations` and actually invoke the method — mirroring your `chatCompletion_shouldHitCorrectEndpointAndReturnResponse` test, which is a genuinely good test.

### 4. Plugin how-to doc not updated

`src/main/resources/doc/io.kestra.plugin.ai.md` lists all available providers under `## Authentication` and doesn't include `DockerModel`. Updating it is required when adding a provider.

---

### Minor

- **Mixed assertion frameworks** — `DockerModelTest` static-imports both `assertj...assertThat` and `hamcrest.MatcherAssert.assertThat` (plus `Matchers.*`), relying on overload resolution to disambiguate. The rest of the package is AssertJ-only. Please standardise on AssertJ and drop the wildcard static imports.
- **`@DockerModelRunnerTest`** — issue #322 asks for a named annotation; the PR inlines `@EnabledIfEnvironmentVariable(named = "DOCKER_MODEL_RUNNER_AVAILABLE")`. Since that variable is never exported in CI the test never runs, which is only acceptable because the WireMock chat test covers the happy path unconditionally. Fine to keep as-is if you prefer, but worth a line in the PR description.
- **`AGENTS.md`** — add `io.kestra.plugin.ai.provider.DockerModel` to the Key Plugin Classes list.
- **Duplicated `"not-needed"`** literal in the `@Builder.Default` initializer and in `resolveApiKey()` — extract a constant next to `DEFAULT_BASE_URL`.

### Nits

- `CHAT_PATH` is a slightly misleading name — that segment serves chat *and* embeddings. `BASE_PATH` reads better.
- Bare `.orElseThrow()` without a message: pre-existing style across the whole provider package, so not something to fix here.

---

Blocking for merge: **2, 3, 4**, and **1** (the getter change is two lines and lets us avoid adding a hook to a shared base class). Everything else is optional polish. Thanks again!

rkampani:
Pushed commit , containing the apiKey-shadowing fix, imageModel fail-fast, rewritten tests, and doc updates

jymaire:
## QA report — live Docker Model Runner validation

**Edition**: OSS | **Docker tag**: `v1.3.30` | **plugin-ai**: `0e39fde` (PR #323) | **plugin-docker**: `16d66b6` ([#123](https://github.com/kestra-io/plugin-docker/pull/123))
**Instance**: http://0k2m9.kestra.docker.localhost:1355/ui/ — left running for manual review (`docker rm -f kestra-0k2m9` to stop, `docker rm -f dmr-qa` for the DMR container)

@rkampani — this closes the gap you flagged: **no M-series Mac was needed.** Docker Model Runner runs fine on Linux/Docker Engine as a standalone container (`docker/model-runner:latest`, `-p 12434:12434 -v <vol>:/models`), so this run exercised a **real DMR** — real HTTP round trips, real CPU inference with `ai/smollm2` (361M params, 256 MiB) and `ai/nomic-embed-text-v1.5`. No GPU on this host.

**The good news: the provider itself works.** Chat, embeddings, RAG, the `apiKey` default, and the shared-parent refactor all behave correctly against a live DMR. **Two things block merge**, both in documentation/UX rather than the core logic — and one of them is in plugin-docker, not here.

### Summary

| # | Flow | Result | Verdict |
|---|---|---|---|
| 1 | `f1_dmr_chat_ok` — chat, gateway baseUrl | ✅ SUCCESS 1.26s | Core of the PR works |
| 2 | `f2_example1_verbatim` — **@Example 1 as written** | ❌ FAILED 0.09s | Blocked by plugin-docker `Pull` (#123) |
| 3 | `f3_role_key_isolated` — `role: user`, valid baseUrl | ❌ FAILED 0.08s | **`role` is not a `ChatMessage` field** |
| 4 | `f4_default_baseurl_isolated` — default baseUrl | ✅ SUCCESS 0.10s | Passes *only* under host networking |
| 5 | `f5_example2_verbatim` — **@Example 2 as written** | ❌ FAILED 0.08s | Same `role` defect |
| 6 | `f6_apikey_explicit` — explicit `apiKey` | ✅ SUCCESS 0.23s | Shadowing fix works functionally |
| 7 | `f7_rag_embeddings` — embeddings + RAG | ✅ SUCCESS 1.49s | `embeddingModel()` works |
| 8 | `f8_image_failfast` — `ImageGeneration`, chat model | ⏱️ KO (timeout) 60s ×2 | Routing correct, **no fail-fast** |
| 9 | `f9_nonreg_openaicompliant` — OpenAI + DeepSeek | ✅ SUCCESS 1.29s | **Non-regression clean** |
| 10 | `f10_desktop_alias_isolated` — Desktop alias | ❌ FAILED 0.33s | Fails with a bare `null` message |

---

### 🔴 1. Both `@Example` flows use `role: user`, which is not a valid field

This is the blocker for this PR. `ChatMessage` has no `role` field:

```
ERROR java.lang.IllegalArgumentException: Unrecognized field "role"
  (class io.kestra.plugin.ai.domain.ChatMessage), not marked as ignorable
  (3 known properties: "content", "type", "contentBlocks")
```

`DockerModel.java:70` and `:95` are the **only two** `role: user` occurrences in the whole repo — the other 25 examples all use `type: USER`. Worse, the flow **saves without complaint** (`POST /flows` → 200) and only fails at execution, so anyone copy-pasting the documented example gets a flow that looks fine and then breaks.

```diff
  messages:
-   - role: user
+   - type: USER
      content: "{{ inputs.prompt }}"
```

Both examples need this. `f3` and `f5` prove it in isolation — `f3` is byte-identical to the passing `f1` except for that one key.

### 🔴 2. `docker.model.Pull` (plugin-docker #123) can't pull from DMR — breaks Example 1

`f2` runs your Example 1 verbatim and dies on the very first task:

```
ERROR DMR /models/create returned HTTP 400
```

`Pull.java` sends `{"fromImage": "<model>"}`. Probed directly against the DMR API:

| Request | Result |
|---|---|
| body `{"fromImage": "ai/smollm2"}` ← current code | **HTTP 400 `Invalid model reference`** |
| query `?fromImage=ai/smollm2` | HTTP 400 `Invalid model reference` |
| body `{"from": "ai/smollm2"}` | **HTTP 200 `Model pulled successfully`** |

The key is `from`, not `fromImage`. This belongs on [plugin-docker#123](https://github.com/kestra-io/plugin-docker/pull/123), but it's what makes this PR's flagship example unusable end-to-end, which is exactly the joint check @fdelbrayelle asked for. Verified against `docker/model-runner:latest` standalone — worth a second look on Docker Desktop in case its DMR build is more lenient.

### 🟡 3. The default `baseUrl` only works under host networking

`f4` (no `baseUrl`, so `http://localhost:12434/engines/v1` applies) **passed** — but only because `kestra-core-run` starts the container with `--network host`. From a normal bridge-networked container:

```
bridge  localhost:12434  -> 000   (unreachable)
bridge  172.17.0.1:12434 -> 200
```

Most Kestra deployments are bridge-networked docker-compose, where `localhost` is the Kestra container itself. The `@Schema` description does document the Engine and Desktop variants, so nothing is *wrong* — but the **default** is the one value that cannot work in the common containerized case. Worth considering whether the default should be the gateway form, or at least whether the docs should lead with the container variants rather than list the host default first.

### 🟡 4. The Docker Desktop alias fails with a bare `null`

`f10` isolates Example 2's `baseUrl` (`model-runner.docker.internal`) with a valid `type: USER`, on Docker Engine:

```
ERROR null
```

That's the entire message. Expected on Engine — the alias is Desktop-only and the docs say so — but a user who copies Example 2 onto Engine gets no hint about what went wrong. A wrapped error naming the unresolvable host would save real debugging time.

### 🟡 5. `imageModel()` routes correctly but does not fail fast

Good news first: **the Diffusers routing works.** DMR logged `Downloading diffusers from Docker Hub... version=v0.1.0-...` on each attempt, which proves the `/engines/v1` → `/engines/diffusers/v1` path replacement lands on the right endpoint.

But `f8` passes a chat model (`ai/smollm2`) — not a diffuser-tagged model — and instead of failing fast it blocks for a full minute:

```
generate  FAILED  PT1M0.035103S
ERROR request timed out
```

Reproduced twice. Nothing validates that the model is diffuser-capable before the call, so the user waits 60s for an opaque timeout. Marked **KO (timeout)** per the 30s QA rule.

One caveat in fairness: on this CPU-only host the diffusers backend download never completed (it restarted on each attempt), so the 60s is dominated by that bootstrap and I could not observe what error surfaces once the backend is cached. The absence of a pre-flight model check is the part I can state with confidence.

### ✅ 6. Non-regression on the shared parent is clean

This PR changes `OpenAICompliantProvider` to read `getApiKey()` instead of the `this.apiKey` field, which puts **every** subclass on a new code path. The real subclasses are `OpenAI`, `DeepSeek`, `DockerModel`, `HuggingFace`. `f9` drives `OpenAI` and `DeepSeek` against DMR's OpenAI-compatible endpoint (no paid key needed) — both SUCCESS. `DeepSeek` specifically matters because it shadows `baseUrl` the same way, and it still resolves correctly through `getBaseUrl()`.

`apiKey` handling is correct in all three shapes: omitted → `not-needed` default applies (`f1`, `f4`), explicitly set → honoured (`f6`). The generated schema is clean, as predicted in review:

```
### apiKey    Required: ❌   Default: `not-needed`
### baseUrl   Required: ❌   Default: `http://localhost:12434/engines/v1`
```

No duplicated property, no spurious `required` entry.

### ℹ️ 7. Review point #1 is only half applied

For the record, since the commit message says "apiKey-shadowing fix":

- ✅ parent now reads `getApiKey()` — this is what makes the child's default actually take effect
- ✅ `DockerModel.resolveApiKey(...)` override removed
- ❌ `DockerModel` still re-declares `private Property<String> apiKey` (line 127) — the shadow is still there
- ❌ `protected final String resolveApiKey(RunContext)` still sits on the shared parent, the new API surface the review asked to drop

No user-visible impact today — schema is clean and flows work, matching the blast-radius analysis in the review — but the parent's `@NotNull apiKey` remains permanently null behind the shadow. Your call whether to finish the cleanup or accept it.

### ℹ️ 8. Unrelated doc nit in a file this PR touches

`io.kestra.plugin.ai.md` advertises "`OpenAICompliantProvider` for any OpenAI-compatible endpoint", but that class is `abstract` and a flow using it is rejected:

```
Invalid entity: Unrecognized field "apiKey" ... 5 known properties
```

Pre-existing, not introduced here — flagging only because this PR edits that line.

---

### Evidence

<details>
<summary>Flow YAML — f1 (core, passing)</summary>

```yaml
id: f1_dmr_chat_ok
namespace: qa.dmr
tasks:
  - id: ask
    type: io.kestra.plugin.ai.completion.ChatCompletion
    provider:
      type: io.kestra.plugin.ai.provider.DockerModel
      modelName: ai/smollm2
      baseUrl: http://172.17.0.1:12434/engines/v1
    messages:
      - type: USER
        content: What is 2 + 2? Reply with just the number.
```
</details>

**f1 outputs** — real inference, token accounting and finish reason all populated:
`textOutput: "2."`, `tokenUsage: {input: 45, output: 3, total: 48}`, `finishReason: STOP`, `requestDuration: 863`

<details>
<summary>Flow YAML — f7 (embeddings + RAG, passing)</summary>

```yaml
id: f7_rag_embeddings
namespace: qa.dmr
tasks:
  - id: ingest
    type: io.kestra.plugin.ai.rag.IngestDocument
    provider:
      type: io.kestra.plugin.ai.provider.DockerModel
      modelName: ai/nomic-embed-text-v1.5
      baseUrl: http://172.17.0.1:12434/engines/v1
    embeddings:
      type: io.kestra.plugin.ai.embeddings.KestraKVStore
    drop: true
    fromDocuments:
      - content: The Kestra QA mascot is a purple otter named Pim.
      - content: Kestra was founded in 2019 and is an orchestration platform.

  - id: chat_with_rag
    type: io.kestra.plugin.ai.rag.ChatCompletion
    chatProvider:
      type: io.kestra.plugin.ai.provider.DockerModel
      modelName: ai/smollm2
      baseUrl: http://172.17.0.1:12434/engines/v1
    embeddingProvider:
      type: io.kestra.plugin.ai.provider.DockerModel
      modelName: ai/nomic-embed-text-v1.5
      baseUrl: http://172.17.0.1:12434/engines/v1
    embeddings:
      type: io.kestra.plugin.ai.embeddings.KestraKVStore
    systemMessage: Answer using only the provided context.
    prompt: What is the name of the Kestra QA mascot?
```
</details>

**f7 Gantt** — `ingest` SUCCESS 0.63s → `chat_with_rag` SUCCESS 0.73s (total 1.49s)
**f7 outputs** — `ingestedDocuments: 2`, `totalTokenCount: 31`, then the RAG answer `"The name of the Kestra QA mascot is Pim."` with both source chunks returned. Retrieval grounded the answer correctly.

<details>
<summary>Flow YAML — f9 (non-regression on the refactored parent)</summary>

```yaml
id: f9_nonreg_openaicompliant
namespace: qa.dmr
tasks:
  - id: chat_via_openai_subclass
    type: io.kestra.plugin.ai.completion.ChatCompletion
    provider:
      type: io.kestra.plugin.ai.provider.OpenAI
      modelName: ai/smollm2
      baseUrl: http://172.17.0.1:12434/engines/v1
      apiKey: not-needed
    messages:
      - type: USER
        content: Reply with the single word OPENAI.

  - id: chat_via_deepseek_subclass
    type: io.kestra.plugin.ai.completion.ChatCompletion
    provider:
      type: io.kestra.plugin.ai.provider.DeepSeek
      modelName: ai/smollm2
      baseUrl: http://172.17.0.1:12434/engines/v1
      apiKey: not-needed
    messages:
      - type: USER
        content: Reply with the single word DEEPSEEK.
```
</details>

**f9 Gantt** — `chat_via_openai_subclass` SUCCESS 0.91s, `chat_via_deepseek_subclass` SUCCESS 0.29s. (`smollm2` is a 361M model, so the *content* of its replies is nonsense — irrelevant here; what matters is that both subclasses resolved credentials and base URL and completed a real round trip.)

**Not covered**: real image generation with a diffuser model (`ai/stable-diffusion` is multi-GB and this host is CPU-only — the diffusers backend never finished bootstrapping); providers requiring paid API keys (covered keylessly via DMR instead); topology/artifact views (no `ui/` changes in this PR, and the feature needs Kestra ≥ 2.0).

**Environment note**: building `plugin-ai` requires **JDK 21** — Lombok fails on JDK 25 with a cascade of `package X does not exist` errors. Not a PR issue, just a heads-up if you hit it.

---

### To merge

1. `role: user` → `type: USER` in both examples (**this PR**)
2. `fromImage` → `from` in `Pull.java` ([plugin-docker#123](https://github.com/kestra-io/plugin-docker/pull/123))

Then Example 1 should run end-to-end. Items 3–5 are judgement calls I'd be happy to see addressed here or in a follow-up; 6–8 need no action.

rkampani:
1. 1he role: user typo — you're right, and it's a bad one.  Fixing both examples to type: USER now — this is the one thing in your report that's squarely in this PR's scope and unambiguous.
2.  Pull's fromImage/from mismatch — that's plugin-docker#123, not this repo. Out of scope — belongs to plugin-docker#123, no change here
3.  the default baseUrl under bridge networking — real finding, but it's Docker networking, not a plugin bug. I don't want to change the default as part of this PR — that'd break the host-networking case for anyone running Kestra
outside a container, and I'd rather not trade one broken default for another without more input. Reordering the schema description to lead with the container variants is cheap if you think it's worth it; I'd file that as a docs
follow-up rather than hold this PR on it.
4. the bare null on the Desktop alias — this deserves a real fix, just not a rushed one. That's almost certainly an unwrapped exception from somewhere in the HTTP client, and wrapping it properly means catching around model
construction without swallowing errors that should stay loud. I'd rather do that right in a follow-up than bolt on a quick try/catch here.
5.  no pre-flight check on diffuser models — same reasoning. There's no documented DMR endpoint for "is this model diffuser-capable," so any check I write now is guessing at tag conventions. And by your own account the diffusers
backend never finished bootstrapping on your host, so we don't even know what error surfaces once it's warm. I'd rather wait for that signal than ship a guess.

7. is where I want to push back. Two things got marked ❌, and I only agree with one of them.
 
   Pushing back: the apiKey field shadow is intentional (matches the original ask, mirrors DeepSeek's baseUrl pattern); resolveApiKey staying protected final is required because  DockerModel.imageModel() calls it — making it private would force duplicated logic the review itself flagged elsewhere
   
6 and 8 — agreed, no action needed.

_Pushing the role: user fix now. For 3–5, since you've got the live DMR environment already running, I'd genuinely rather you drive those as follow-ups than have me guess blind on my end._

jymaire:
@rkampani  didn't wanted to update your branch so I opened a PR on your fork with the follow ups https://github.com/kestra-io/plugin-openai/pull/188

I will merge plugin-docker PR to unblock the PR

rkampani:
> @rkampani didn't wanted to update your branch so I opened a PR on your fork with the follow ups [kestra-io/plugin-openai#188](https://github.com/kestra-io/plugin-openai/pull/188)
> 
> I will merge plugin-docker PR to unblock the PR

Thank you! I merged your PR—really appreciate your help with this.

Please let me know if there are any remaining items that need to be addressed. Otherwise, I think we're good to move forward with the merge.

MilosPaunovic:
@Malaydewangan09 Give this a look if you have the capacity for it. 🙏🏻

Malaydewangan09:
Yep, this PR is on my list to review. Thanks!

Malaydewangan09:
Reviewed again:

Nothing blocks merge. Two optional cleanups if you want:
- Extract the `"not-needed"` literal (DockerModel.java:155) into a constant next to `DEFAULT_BASE_URL`.
- `imageModel()` reads `this.baseUrl` (line 176) while the rest uses `getBaseUrl()`.

Both cosmetic. LGTM.

Malaydewangan09 (approved):

Malaydewangan09 (approved):

PR #414, opened 2026-09-01
fix(tool): report Kestra API auth failures as such instead of a missing flow

### LINKED ISSUE #404: KestraFlow tool reports an authentication failure as `Flow not found` and sends an unauthenticated request when no credentials are found
When the `KestraFlow` tool is used and no credentials can be resolved, the request is still sent to the Kestra API without authentication. The rejection is then reported to the agent as `Flow not found: <namespace>.<flowId>`, even when the flow exists. Nothing in the message points at authentication, so the flow ID and the namespace are the first things checked.

Currently this happens when no `auth` block is set on the tool. With an `auth` block present and no usable credentials, the failure is reported as a missing authentication method, so the same condition is handled two different ways.

Minimal example, on an instance where authentication is enabled and no credentials are configured for the tool:

```yaml
id: kestra_flow_tool_auth
namespace: company.team

tasks:
  - id: agent
    type: io.kestra.plugin.ai.agent.AIAgent
    prompt: Execute the flow hello-world in the namespace company.team
    provider:
      type: io.kestra.plugin.ai.provider.GoogleGemini
      modelName: gemini-3.5-flash
      apiKey: "{{ secret('GEMINI_API_KEY') }}"
    tools:
      - type: io.kestra.plugin.ai.tool.KestraFlow
```

`Flow not found: company.team.hello-world` is reported, even though `company.team.hello-world` exists.

## To do

- An authentication or authorization failure should be reported as such, with a message that names authentication rather than a missing flow. Only a genuine 404 should be reported as a flow that cannot be found.
- When automatic credential retrieval is enabled and no credentials can be found, the tool should fail at that point instead of sending an unauthenticated request. The same behavior should apply whether or not an `auth` block is set.

brian-mulier-p:
Closes https://github.com/kestra-io/plugin-ai/issues/404.

- A 401 or a 403 from the Kestra API was reported as `Flow not found`, so a wrong or unauthorized credential sent the user looking for a flow that was there all along. Each status is now named for what it is, and only a genuine 404 is reported as a missing flow. The mapping covers the three places the API is called: resolving a flow defined on the tool, resolving one the LLM asked for, and creating the execution.
- Without an `auth` block the tool fell back to an unauthenticated request when automatic credential retrieval found nothing, producing that same misleading 401. It now fails at that point, the way it already did when an `auth` block was set. `auto: false` with no credentials keeps throwing, unchanged.
- A `ToolArgumentsException` raised for a missing input was being wrapped into a generic `ToolExecutionException`; it is now rethrown as-is, so the right handler and observability listener fire. (The agent's `toolArgumentsErrorHandler` still rethrows rather than returning the error to the LLM, so this does not by itself enable a retry — see the review discussion.)
- `kestraUrl` is untouched — same name, same place, same meaning. Its resolution moves into a helper that rejects a blank `{{ kestra.url }}` rather than building a client with an empty base path. **Not a breaking change.**

Review follow-up:
- The auth messages now actually reach the agent: langchain4j reports the root cause of what an executor throws, so the two executor sites no longer chain the `ApiException` (it is logged at warn instead), and the tests pin the surfaced message.
- `auth.auto: false` with no credentials still throws, and cannot be turned into an anonymous call from here: the SDK's `KestraClientBuilder` starts on `Auth.BASIC` and exposes no setter for `Auth.NONE`, so `build()` without credentials would send `Authorization: Basic base64("null:null")` rather than no header. Calling an unsecured API needs a `noAuth()` on that builder in `client-sdk` first.
- Both doc examples gained the `auth` block they need on an instance without a default authentication, and `pluginDefaults` is gone from every example in this plugin since it no longer exists after Kestra 2.0.
- An `ApiException` with status 0 (no HTTP response at all) is reported as an unreachable API with the underlying message, not as `status 0`, and any unmapped status now carries the API's own message, which this SDK exposes as the response body.
- An unset tenant was serialized as `""`, producing `/api/v1//flows/...` paths; it now falls back to the flow's tenant and then to `main`.

```yaml
id: agent_calling_a_flow
namespace: company.ai

tasks:
  - id: agent
    type: io.kestra.plugin.ai.agent.AIAgent
    provider:
      type: io.kestra.plugin.ai.provider.GoogleGemini
      apiKey: "{{ kv('GEMINI_API_KEY') }}"
      modelName: gemini-2.5-flash
    prompt: Run the hello-world flow and give me its execution id.
    tools:
      - type: io.kestra.plugin.ai.tool.KestraFlow
        namespace: company.team
        flowId: hello-world
        kestraUrl: http://localhost:8080
        auth:
          apiToken: "{{ kv('KESTRA_API_TOKEN') }}"
```

With a wrong `apiToken` the agent now reports `Authentication failed when calling the Kestra API for the flow 'hello-world' in the namespace 'company.team'. Check the credentials set in the auth property of the tool.` instead of `Flow not found: company.team.hello-world`. Dropping the whole `auth` block on an instance with no `kestra.tasks.sdk.authentication` configured fails before any request, with `No authentication method provided`.

Fourteen tests in `KestraFlowTest` cover the three failure sites, the fail-fast on missing credentials, the arguments exception, an unreachable API and an unmapped status.

### DIFF
--- a/src/main/java/io/kestra/plugin/ai/tool/KestraFlow.java
+++ b/src/main/java/io/kestra/plugin/ai/tool/KestraFlow.java
@@ -3,6 +3,7 @@ package io.kestra.plugin.ai.tool;
 import java.time.OffsetDateTime;
 import java.time.ZonedDateTime;
 import java.util.*;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
@@ -17,6 +18,7 @@ import io.kestra.core.models.property.Property;
 import io.kestra.core.runners.RunContext;
 import io.kestra.core.runners.SDK;
 import io.kestra.core.serializers.JacksonMapper;
+import io.kestra.core.tenant.TenantService;
 import io.kestra.core.serializers.ListOrMapOfLabelDeserializer;
 import io.kestra.core.serializers.ListOrMapOfLabelSerializer;
 import io.kestra.core.utils.IdUtils;
@@ -36,6 +38,7 @@ import com.fasterxml.jackson.core.type.TypeReference;
 
 import dev.langchain4j.agent.tool.ToolExecutionRequest;
 import dev.langchain4j.agent.tool.ToolSpecification;
+import dev.langchain4j.exception.LangChain4jException;
 import dev.langchain4j.exception.ToolArgumentsException;
 import dev.langchain4j.exception.ToolExecutionException;
 import dev.langchain4j.model.chat.request.json.JsonArraySchema;
@@ -63,6 +66,12 @@ import static io.kestra.core.utils.Rethrow.throwFunction;
                 """
                     id: agent_calling_flows_explicitly
                     namespace: company.ai
+
+                    pluginDefaults:
+                      - type: io.kestra.plugin.ai.tool.KestraFlow
+                        values:
+                          auth:
+                            apiToken: "{{ secret('KESTRA_API_TOKEN') }}"
 
                     inputs:
                       - id: use_case
@@ -168,7 +177,9 @@ import static io.kestra.core.utils.Rethrow.throwFunction;
                           modelName: gemini-3.5-flash-lite
                           apiKey: "{{ secret('GEMINI_API_KEY') }}"
                         tools:
-                          - type: io.kestra.plugin.ai.tool.KestraFlow"""
+                          - type: io.kestra.plugin.ai.tool.KestraFlow
+                            auth:
+                              apiToken: "{{ secret('KESTRA_API_TOKEN') }}\""""
             }
         ),
     }
@@ -185,6 +196,9 @@ public class KestraFlow extends ToolProvider {
     private static final String TOOL_LLM_DESCRIPTION = """
         This tool executes a Kestra workflow, also called a flow. This tool will respond with the flow execution information.
         The namespace and the ID of the flow must be passed as tool parameters.""";
+
+    private static final String DEFAULT_URL = "http://localhost:8080";
+    private static final String URL_TEMPLATE = "{{ kestra.url }}";
 
     @Schema(
         title = "Description of the flow if not already provided inside the flow itself",
@@ -276,18 +290,33 @@ public class KestraFlow extends ToolProvider {
         return Optional.empty();
     }
 
+    private String resolveUrl(RunContext runContext) throws IllegalVariableEvaluationException {
+        String rUrl = runContext.render(kestraUrl).as(String.class)
+            .filter(Predicate.not(String::isBlank))
+            .orElseGet(() -> configuredUrl(runContext));
+
+        return rUrl.trim().replaceAll("/+$", "");
+    }
+
+    private static String configuredUrl(RunContext runContext) {
+        try {
+            String rUrl = runContext.render(URL_TEMPLATE);
+            return rUrl == null || rUrl.isBlank() ? DEFAULT_URL : rUrl;
+        } catch (IllegalVariableEvaluationException e) {
+            return DEFAULT_URL;
+        }
+    }
+
+    private static IllegalArgumentException noAuthentication() {
+        return new IllegalArgumentException(
+            "No authentication method provided. Set the `auth` property of the tool, or configure a default one with the `kestra.tasks.sdk.authentication` properties."
+        );
+    }
+
     private KestraClient kestraClient(RunContext runContext) throws IllegalVariableEvaluationException {
-        String rKestraUrl = runContext.render(kestraUrl).as(String.class)
-            .orElseGet(() -> {
-                try {
-                    return runContext.render("{{ kestra.url }}");
-                } catch (IllegalVariableEvaluationException e) {
-                    return "http://localhost:8080";
-                }
-            });
-        String normalizedUrl = rKestraUrl.trim().replaceAll("/+$", "");
         var builder = KestraClient.builder();
-        builder.url(normalizedUrl);
+        builder.url(resolveUrl(runContext));
+
         if (auth != null) {
             String rApiToken = runContext.render(auth.apiToken).as(String.class).orElse(null);
             if (rApiToken != null) {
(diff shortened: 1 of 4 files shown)


jymaire:
## Code review

Reviewed `KestraFlow.java` and `KestraFlowTest.java` at `7ae1d61`. One blocker, then a few things worth deciding before merge.

### 1. The new auth messages never reach the user (HIGH) — `KestraFlow.java:561`, `:476`

The core fix doesn't actually surface. Both executor sites throw:

```java
throw new ToolExecutionException(apiFailureMessage(...), e);
```

LangChain4j's `ToolService.executeWithErrorHandling` catches this and calls the configured handler with `getCause(e)` — i.e. the chained `ApiException`, not the exception we threw (verified in `langchain4j-1.19.0`, `ToolService.getCause`). `AIAgent` (and `completion.ChatCompletion`, `rag.ChatCompletion`, `tool.AIAgent`) install a `toolExecutionErrorHandler` that does `throw new ToolExecutionException(error)`, and `LangChain4jException(Throwable)` sets its message to `cause.getMessage()`.

Net result: "Authentication failed when calling the Kestra API for the flow …" is discarded and the SDK's raw 401 message is shown instead — exactly the situation #404 is about.

The `tool()` site (`:393`, `IllegalArgumentException`) is outside this pipeline, which is why the manual QA on a tool-defined flow looked correct. The new tests call `executor.execute(...)` directly and bypass `executeWithErrorHandling`, so they can't catch this either.

Fix: either don't chain the `ApiException` as the cause at those two sites (throw `new ToolExecutionException(message)` and log the cause), or change the agent handlers to preserve the thrown exception's message.

### 2. No way left to call an unsecured Kestra API (MEDIUM) — `:319`

With `auth` omitted the tool throws instead of building an unauthenticated client, and there's no opt-in for anonymous access: `auth: {auto: false}` with no credentials also throws (`:316`), and `auth: {}` throws when `kestra.tasks.sdk.authentication` is unset. Any instance whose API accepts unauthenticated calls (basic auth disabled, a proxy injecting auth, local dev) loses the tool with "No authentication method provided". Worth keeping `auto: false` + no credentials as the explicit "no auth" escape hatch.

### 3. Both `@Example` snippets are broken as written (MEDIUM) — `:92`, `:173`

Neither `agent_calling_flows_explicitly` nor `agent_calling_flows_implicitly` defines an `auth` block, so on any instance without `kestra.tasks.sdk.authentication` configured they now fail at `tool()`. These are the snippets rendered in the plugin docs and copy-pasted by users — they need an `auth:` block alongside the `kestraUrl` → `auth.url` move.

### 4. `default ->` reports "status 0" and swallows the real cause (LOW) — `:328`

`ApiException` also carries code `0` when there's no HTTP response at all (unreachable host, wrong port, TLS failure) — precisely the misconfiguration a freshly-introduced `auth.url` invites. The user gets "The Kestra API returned the status 0 for the flow 'x' in the namespace 'y'.", as misleading as the "Flow not found" this PR removes. Special-case `code == 0` and include `e.getMessage()`.

### 5. `kestraUrl` removed with no deprecation alias (LOW) — `:625`

Every existing flow using `kestraUrl:` becomes invalid on upgrade (unknown property). The break is intentional and flagged in the title, but a `@Deprecated` field forwarding to `auth.url` for one release costs little.

Consistency point: the sibling `io.kestra.plugin.ai.tool.KestraTask` wraps `io.kestra.plugin.kestra.AbstractKestraTask`, which still uses `kestraUrl` + an `auth` block *without* a `url` — so an agent using both tools now needs two different connection layouts.

### 6. Rethrowing `ToolArgumentsException` doesn't give the LLM a retry (LOW) — `:541`

`AIAgent.java:775` installs a `toolArgumentsErrorHandler` that throws `ToolArgumentsException(error)` rather than returning `ToolErrorHandlerResult.text(...)`, and langchain4j's default arguments handler also rethrows. So both before and after this change the agent run aborts; the only observable differences are which observability listener fires and the log line. If letting the LLM retry with different arguments is the goal, the handlers in `AIAgent` / `completion.ChatCompletion` / `rag.ChatCompletion` need to return a `ToolErrorHandlerResult` instead of throwing.

### Verified as correct

- `catch (LangChain4jException e) { throw e; }` before `catch (Exception e)` compiles and behaves as intended — `ToolExecutionException` and `ToolArgumentsException` both extend `LangChain4jException` directly in langchain4j 1.19.0.
- `resolveUrl` handles a null `auth` and a null/blank `auth.url`; passing a null `Property` to `runContext.render` was already the pre-existing pattern for `kestraUrl`.
- No remaining `kestraUrl` references for `KestraFlow` anywhere in the repo (the `KestraTaskTest` hit is `AbstractKestraTask`, unrelated).
- `src/test/resources/application.yml` sets no SDK authentication, so `shouldFailBeforeCallingTheApiWhenNoCredentialsCanBeRetrieved` is deterministic.

src/test/java/io/kestra/plugin/ai/tool/KestraFlowTest.java
```diff
@@ -354,4 +369,110 @@ void helloWorldFromLLM() throws Exception {
...
+            .build();
+    }
+
+    private ToolExecutor executorOf(KestraFlow tool) throws Exception {
+        return tool.tool(runContextFactory.of(), Map.of()).values().iterator().next();
+    }
+
+    @Test
+    void shouldReportAnAuthenticationFailureWhenTheApiRejectsTheCredentials() {
+        stubStatuses.put("/api/v1//flows/company.team/hello-world", 401);
```
jymaire: strange the double /

src/main/java/io/kestra/plugin/ai/tool/KestraFlow.java
```diff
@@ -299,13 +338,23 @@ private KestraClient kestraClient(RunContext runContext) throws IllegalVariableE
...
+    }
+
+    /** A 401 or a 403 from the API is a credentials problem, and reporting it as a missing flow sends the user looking in the wrong place. */
+    private static String apiFailureMessage(ApiException e, String namespace, String flowId) {
+        return switch (e.getCode()) {
+            case 401 -> "Authentication failed when calling the Kestra API for the flow '%s' in the namespace '%s'. Check the credentials set in the `auth` property of the tool.".formatted(flowId, namespace);
+            case 403 -> "Not authorized to access the flow '%s' in the namespace '%s'. Check the permissions of the credentials set in the `auth` property of the tool.".formatted(flowId, namespace);
+            case 404 -> "Unable to find the flow '%s' in the namespace '%s'.".formatted(flowId, namespace);
+            case 0 -> "The Kestra API could not be reached for the flow '%s' in the namespace '%s': %s".formatted(flowId, namespace, e.getMessage());
+            default -> "The Kestra API returned the status %d for the flow '%s' in the namespace '%s'.".formatted(e.getCode(), flowId, namespace);
```
jymaire: maybe add `e.getMessage` in the default case so we get more information?

src/test/java/io/kestra/plugin/ai/tool/KestraFlowTest.java
```diff
@@ -117,10 +132,10 @@ private String executionJson(String id, String namespace, String flowId) {
 
     @Test
     void helloWorld() throws Exception {
-        // The SDK uses an empty tenant string, producing a double-slash path: /api/v1//flows/...
-        stubFlowResponses.put("/api/v1//flows/company.team/hello-world",
+        // The SDK uses an empty tenant string, producing a double-slash path: /api/v1/main/flows/...
```
jymaire: I didn't see this comment in previous review when I pointed out the double slash. So either the SDK does add double slash so my first comment was bad, either the comment is out of date

src/main/java/io/kestra/plugin/ai/tool/KestraFlow.java
```diff
@@ -276,18 +300,33 @@ private Optional<KestraClient> tryAutoAuth(KestraClient.KestraClientBuilder buil
...
+            String rUrl = runContext.render(URL_TEMPLATE);
+            return rUrl == null || rUrl.isBlank() ? DEFAULT_URL : rUrl;
+        } catch (IllegalVariableEvaluationException e) {
+            return DEFAULT_URL;
+        }
+    }
+
+    private static IllegalArgumentException noAuthentication() {
+        return new IllegalArgumentException(
+            "No authentication method provided. Set the `auth` property of the tool, configure a default one with the `kestra.tasks.sdk.authentication` properties, or set `auth.auto: false` to call an unsecured Kestra API."
```
jymaire: maybe add the `auth.auto: false` part in some doc ?

jymaire (approved):

PR #417, opened 2026-09-03
fix(rag): apply IngestDocument top-level metadata to every document

### LINKED ISSUE #395: [Bash2.0] IngestDocument top-level metadata is silently dropped
## Describe the issue

`rag.IngestDocument` exposes a top-level `metadata` property, but it is never applied to ingested documents. The property is declared and rendered nowhere in `run()`; only the per-document `metadata` on inline documents (`fromDocuments[].metadata`) is attached. The field's own description admits this ("this top-level property is not currently applied to ingested documents").

The result is silent data loss: a user who sets top-level `metadata` expecting it on every ingested document gets nothing, with no error or warning. This also means documents ingested from `fromPath`, `fromInternalURIs`, or `fromExternalURLs` (which have no per-document metadata field) cannot be given metadata at all.

## Steps to reproduce

1. Ingest documents with top-level metadata set:

   ```yaml
   id: rag_ingest
   namespace: company.ai
   tasks:
     - id: ingest
       type: io.kestra.plugin.ai.rag.IngestDocument
       provider:
         type: io.kestra.plugin.ai.provider.GoogleGemini
         apiKey: "{{ secret('GEMINI_API_KEY') }}"
         modelName: text-embedding-004
       embeddings:
         type: io.kestra.plugin.ai.embeddings.KestraKVStore
       fromExternalURLs:
         - https://raw.githubusercontent.com/kestra-io/docs/main/README.md
       metadata:
         source: "manual-run"
         team: "data"
   ```

2. Run, then query the embedding store and inspect the stored documents' metadata.

## Actual behaviour

The ingested documents carry no `source`/`team` metadata. The top-level `metadata` field (`IngestDocument.java:152`) is never read in `run()`; only `inlineDocument.metadata` (`IngestDocument.java:211`) is applied, and only for `fromDocuments` entries.

## Expected behaviour

Either apply the top-level `metadata` to every ingested document (all sources), or remove the property so it does not silently accept input that is discarded.

jymaire:
## Summary

`rag.IngestDocument` exposed a top-level `metadata` property that was never read in `run()` — it was silently dropped, and the three non-inline sources (`fromPath`, `fromInternalURIs`, `fromExternalURLs`) had no way at all to attach metadata to ingested documents.

- Render `metadata` once at the start of `run()` and merge it into every document from all four sources, at `batch.add` time so a configured `documentSplitter` propagates it to each segment.
- Existing keys always win: inline `fromDocuments[].metadata` and loader-injected keys (`file_name`, `absolute_directory_path`) are never overwritten.
- Widened the property to `Property<Map<String, Object>>` (it was `Map<String, String>`) so numeric metadata is usable; values are validated against the types LangChain4j `Metadata` accepts, up front, so an unsupported value fails with a message naming the key instead of aborting mid-ingestion.
- `null` values are skipped; unset or empty `metadata` is a no-op.
- Documented the property in the package-level how-to doc (`doc/io.kestra.plugin.ai.md`) alongside the other `IngestDocument` properties.
- Rewrote the `@Schema` description (it previously admitted the property was not applied) and added an example matching the issue repro.

closes: https://github.com/kestra-io/plugin-ai/issues/395

## User-visible behaviour change (release note)

A flow that set a top-level `metadata` value of an unsupported type (a boolean, a list, a map, or an out-of-range number) previously succeeded, because the property was never read. It now fails at task start with a message naming the offending key. This is the intended consequence of actually applying the property; quote the value to send it as a String.

Also in this PR, `fromInternalURIs` now decodes documents as explicit UTF-8 instead of the JVM default charset. This is a no-op on a JVM whose default is already UTF-8 (the default since JEP 400) but changes decoding on one started with an overridden `-Dfile.encoding`.

## Test plan

- [x] `./gradlew test --tests '*IngestDocumentTest*'` — 11 tests green. The four new metadata tests were confirmed red before the fix (top-level values absent from the stored documents).
- [x] New coverage: top-level metadata on `fromInternalURIs` / `fromExternalURLs` / `fromPath`, inline metadata precedence, loader `file_name` preservation, no-metadata regression guard, unsupported value type.
- [ ] Full suite: only pre-existing environment failures remain (`ImageGenerationTest` binds WireMock to the hardcoded port 8080, occupied locally) — unrelated to this change.

---
*[View as Artifact](https://claude.ai/code/artifact/76d06512-c806-4b20-8660-f44a97c7b20f)*


jymaire:
## QA report — browser-based

**Edition**: OSS | **Docker tag**: `develop` (2.0)
**Instance**: http://0vfr7.kestra.docker.localhost:1355 (left running for manual review — `docker rm -f kestra-0vfr7` to stop)
**Screenshots**: https://claude.ai/code/artifact/3f56a295-03df-45ba-a9c5-ea3e853b9357
**Plugin build**: `plugin-ai 1.16.1-SNAPSHOT` from `fix/ingest-document-top-level-metadata`, deployed to `PLUGINS_DIR` before boot

> QA was planned for EE, but the EE image pull needs a fresh `gcloud` token that had expired. Switched to OSS with the user's approval: `rag.IngestDocument` is OSS plugin code with no EE-specific surface, so the code path exercised is identical.

Embeddings ran fully offline through Docker Model Runner (`ai/nomic-embed-text-v1.5`) into `KestraKVStore`, so no API key was involved. Each scenario writes its own KV key, and metadata was verified by reading the stored segments back.

### Summary

| # | Flow | Status | Verdict |
|---|------|--------|---------|
| 1 | `ingest_external_urls_metadata` | SUCCESS | ✅ issue repro fixed |
| 2 | `ingest_inline_precedence` | SUCCESS | ✅ inline wins, top-level fills gaps |
| 3 | `ingest_frompath_loader_keys` | SUCCESS | ✅ loader keys preserved |
| 4 | `ingest_internal_uris_metadata` | SUCCESS | ✅ previously impossible, now works |
| 5 | `ingest_no_metadata` | SUCCESS | ✅ no regression |
| 6 | `ingest_bad_metadata_type` | FAILED *(expected)* | ✅ clear error, no partial write |
| 7 | `ingest_with_splitter_and_search` | SUCCESS | ✅ no regression |

**7/7 as expected.** No timeouts — the slowest execution was 2.46s.

---

### Flow 1: ingest_external_urls_metadata (✅ SUCCESS) — the issue repro

<details>
<summary>Flow YAML</summary>

```yaml
id: ingest_external_urls_metadata
namespace: qa.rag
tasks:
  - id: ingest
    type: io.kestra.plugin.ai.rag.IngestDocument
    provider:
      type: io.kestra.plugin.ai.provider.DockerModel
      modelName: ai/nomic-embed-text-v1.5
      baseUrl: http://172.17.0.1:12434/engines/v1
    embeddings:
      type: io.kestra.plugin.ai.embeddings.KestraKVStore
      kvName: qa_external
    documentSplitter:
      splitter: RECURSIVE
      maxSegmentSizeInChars: 400
      maxOverlapSizeInChars: 0
    fromExternalURLs:
      - https://raw.githubusercontent.com/kestra-io/docs/main/README.md
    metadata:
      source: "manual-run"
      team: "data"
  - id: read_kv
    type: io.kestra.plugin.core.kv.Get
    namespace: qa.rag
    key: qa_external
```
</details>

**Gantt** ([screenshot](https://claude.ai/code/artifact/3f56a295-03df-45ba-a9c5-ea3e853b9357#ingest_external_urls_metadata-gantt))
| Task | Status | Duration |
|------|--------|----------|
| `ingest` | SUCCESS | 2.18s |
| `read_kv` | SUCCESS | 0.11s |
| **Total** | **SUCCESS** | **2.46s** |

**Logs synthesis** — clean run, no warnings.

**Outputs synthesis** ([screenshot](https://claude.ai/code/artifact/3f56a295-03df-45ba-a9c5-ea3e853b9357#ingest_external_urls_metadata-outputs))
`ingest` returned `ingestedDocuments: 1`, `inputTokenCount: 1317`, `totalTokenCount: 1317`, `embeddingStoreOutputs.kvName: qa_external`.

**Stored metadata** — all 13 segments carry the top-level keys, plus the loader's `url` and the splitter's `index`:
```json
{"index": "2", "source": "manual-run", "team": "data", "url": "https://raw.githubusercontent.com/kestra-io/docs/main/README.md"}
```
On `main` this same flow stored neither `source` nor `team`.

---

### Flow 2: ingest_inline_precedence (✅ SUCCESS)

<details>
<summary>Flow YAML</summary>

```yaml
id: ingest_inline_precedence
namespace: qa.rag
tasks:
  - id: ingest
    type: io.kestra.plugin.ai.rag.IngestDocument
    provider:
      type: io.kestra.plugin.ai.provider.DockerModel
      modelName: ai/nomic-embed-text-v1.5
      baseUrl: http://172.17.0.1:12434/engines/v1
    embeddings:
      type: io.kestra.plugin.ai.embeddings.KestraKVStore
      kvName: qa_inline
    fromDocuments:
      - content: "Inline document A about Kestra orchestration."
        metadata:
          source: "inline-wins"
          doc: "a"
      - content: "Inline document B about metadata precedence."
        metadata:
          doc: "b"
    metadata:
      source: "top-level-should-lose"
      team: "data"
  - id: read_kv
    type: io.kestra.plugin.core.kv.Get
    namespace: qa.rag
    key: qa_inline
```
</details>

**Gantt** ([screenshot](https://claude.ai/code/artifact/3f56a295-03df-45ba-a9c5-ea3e853b9357#ingest_inline_precedence-gantt))
| Task | Status | Duration |
|------|--------|----------|
| `ingest` | SUCCESS | 0.39s |
| `read_kv` | SUCCESS | 0.03s |
| **Total** | **SUCCESS** | **0.55s** |

**Stored metadata** ([outputs screenshot](https://claude.ai/code/artifact/3f56a295-03df-45ba-a9c5-ea3e853b9357#ingest_inline_precedence-outputs))
```json
{"doc": "a", "index": "0", "source": "inline-wins",           "team": "data"}
{"doc": "b", "index": "0", "source": "top-level-should-lose", "team": "data"}
```
Doc A keeps its own `source`; doc B, which declares none, inherits the top-level one. Both inherit `team`. Precedence is exactly as documented.

---

### Flow 3: ingest_frompath_loader_keys (✅ SUCCESS)

<details>
<summary>Flow YAML</summary>

```yaml
id: ingest_frompath_loader_keys
namespace: qa.rag
tasks:
  - id: wdir
    type: io.kestra.plugin.core.flow.WorkingDirectory
    namespaceFiles:
      enabled: true
    tasks:
      - id: ingest
        type: io.kestra.plugin.ai.rag.IngestDocument
        provider:
          type: io.kestra.plugin.ai.provider.DockerModel
          modelName: ai/nomic-embed-text-v1.5
          baseUrl: http://172.17.0.1:12434/engines/v1
        embeddings:
          type: io.kestra.plugin.ai.embeddings.KestraKVStore
          kvName: qa_frompath
        fromPath: docs
        metadata:
          source: "frompath-run"
          team: "data"
  - id: read_kv
    type: io.kestra.plugin.core.kv.Get
    namespace: qa.rag
    key: qa_frompath
```
</details>

**Gantt** ([screenshot](https://claude.ai/code/artifact/3f56a295-03df-45ba-a9c5-ea3e853b9357#ingest_frompath_loader_keys-gantt))
| Task | Status | Duration |
|------|--------|----------|
| `wdir` | SUCCESS | 0.25s |
| `ingest` | SUCCESS | 0.14s |
| `read_kv` | SUCCESS | 0.03s |
| **Total** | **SUCCESS** | **0.33s** |

**Stored metadata** ([outputs screenshot](https://claude.ai/code/artifact/3f56a295-03df-45ba-a9c5-ea3e853b9357#ingest_frompath_loader_keys-outputs))
```json
{"absolute_directory_path": "/tmp/.../docs", "file_name": "alpha.txt", "index": "0", "source": "frompath-run", "team": "data"}
{"absolute_directory_path": "/tmp/.../docs", "file_name": "beta.txt",  "index": "0", "source": "frompath-run", "team": "data"}
```
Loader-injected provenance keys survive the merge — the naive `Metadata.from(...)` replacement the plan warned about would have wiped these.

---

### Flow 4: ingest_internal_uris_metadata (✅ SUCCESS)

<details>
<summary>Flow YAML</summary>

```yaml
id: ingest_internal_uris_metadata
namespace: qa.rag
tasks:
  - id: write_file
    type: io.kestra.plugin.core.storage.Write
    content: "Document stored in Kestra internal storage for the fromInternalURIs QA scenario."
    extension: .txt
  - id: ingest
    type: io.kestra.plugin.ai.rag.IngestDocument
    provider:
      type: io.kestra.plugin.ai.provider.DockerModel
      modelName: ai/nomic-embed-text-v1.5
      baseUrl: http://172.17.0.1:12434/engines/v1
    embeddings:
      type: io.kestra.plugin.ai.embeddings.KestraKVStore
      kvName: qa_internal
    fromInternalURIs:
      - "{{ outputs.write_file.uri }}"
    metadata:
      source: "internal-uri-run"
      team: "data"
  - id: read_kv
    type: io.kestra.plugin.core.kv.Get
    namespace: qa.rag
    key: qa_internal
```
</details>

**Gantt** ([screenshot](https://claude.ai/code/artifact/3f56a295-03df-45ba-a9c5-ea3e853b9357#ingest_internal_uris_metadata-gantt))
| Task | Status | Duration |
|------|--------|----------|
| `write_file` | SUCCESS | 0.03s |
| `ingest` | SUCCESS | 0.13s |
| `read_kv` | SUCCESS | 0.04s |
| **Total** | **SUCCESS** | **0.32s** |

**Stored metadata** ([outputs screenshot](https://claude.ai/code/artifact/3f56a295-03df-45ba-a9c5-ea3e853b9357#ingest_internal_uris_metadata-outputs))
```json
{"index": "0", "source": "internal-uri-run", "team": "data"}
```
This source had no metadata mechanism whatsoever before the PR.

---

### Flow 5: ingest_no_metadata (✅ SUCCESS) — non-regression

<details>
<summary>Flow YAML</summary>

```yaml
id: ingest_no_metadata
namespace: qa.rag
tasks:
  - id: ingest
    type: io.kestra.plugin.ai.rag.IngestDocument
    provider:
      type: io.kestra.plugin.ai.provider.DockerModel
      modelName: ai/nomic-embed-text-v1.5
      baseUrl: http://172.17.0.1:12434/engines/v1
    embeddings:
      type: io.kestra.plugin.ai.embeddings.KestraKVStore
      kvName: qa_nometa
    fromDocuments:
      - content: "Regression guard: no top-level metadata is configured on this task."
        metadata:
          doc: "only-inline"
  - id: read_kv
    type: io.kestra.plugin.core.kv.Get
    namespace: qa.rag
    key: qa_nometa
```
</details>

**Gantt** ([screenshot](https://claude.ai/code/artifact/3f56a295-03df-45ba-a9c5-ea3e853b9357#ingest_no_metadata-gantt))
| Task | Status | Duration |
|------|--------|----------|
| `ingest` | SUCCESS | 0.12s |
| `read_kv` | SUCCESS | 0.04s |
| **Total** | **SUCCESS** | **0.23s** |

**Stored metadata** ([outputs screenshot](https://claude.ai/code/artifact/3f56a295-03df-45ba-a9c5-ea3e853b9357#ingest_no_metadata-outputs))
```json
{"doc": "only-inline", "index": "0"}
```
Only the inline key and the store's own `index`. Nothing leaks in when `metadata` is unset.

---

### Flow 6: ingest_bad_metadata_type (✅ FAILED as designed) — error path

<details>
<summary>Flow YAML</summary>

```yaml
id: ingest_bad_metadata_type
namespace: qa.rag
tasks:
  - id: ingest
    type: io.kestra.plugin.ai.rag.IngestDocument
    provider:
      type: io.kestra.plugin.ai.provider.DockerModel
      modelName: ai/nomic-embed-text-v1.5
      baseUrl: http://172.17.0.1:12434/engines/v1
    embeddings:
      type: io.kestra.plugin.ai.embeddings.KestraKVStore
      kvName: qa_badtype
    fromDocuments:
      - content: "This ingestion must fail before anything is written to the store."
    metadata:
      source: "bad-type-run"
      tags:
        - "alpha"
        - "beta"
```
</details>

**Gantt** ([screenshot](https://claude.ai/code/artifact/3f56a295-03df-45ba-a9c5-ea3e853b9357#ingest_bad_metadata_type-gantt))
| Task | Status | Duration |
|------|--------|----------|
| `ingest` | FAILED | 0.04s |
| **Total** | **FAILED** | **0.08s** |

**Logs synthesis** ([screenshot](https://claude.ai/code/artifact/3f56a295-03df-45ba-a9c5-ea3e853b9357#ingest_bad_metadata_type-logs)) — exactly one ERROR:
```
The metadata key 'tags' has an unsupported value type 'ArrayList' — use one of String, UUID, Integer, Long, Float or Double.
```
The message names the offending key and its type, as the plan required. **The KV key `qa_badtype` was never created** (API returns 404), confirming validation runs before the first flush and no partial ingestion occurs.

---

### Flow 7: ingest_with_splitter_and_search (✅ SUCCESS) — non-regression

<details>
<summary>Flow YAML</summary>

```yaml
id: ingest_with_splitter_and_search
namespace: qa.rag
tasks:
  - id: ingest
    type: io.kestra.plugin.ai.rag.IngestDocument
    provider:
      type: io.kestra.plugin.ai.provider.DockerModel
      modelName: ai/nomic-embed-text-v1.5
      baseUrl: http://172.17.0.1:12434/engines/v1
    embeddings:
      type: io.kestra.plugin.ai.embeddings.KestraKVStore
      kvName: qa_search
    documentSplitter:
      splitter: RECURSIVE
      maxSegmentSizeInChars: 200
      maxOverlapSizeInChars: 0
    fromDocuments:
      - content: "Kestra is an open source orchestration platform. It runs workflows declared in YAML. Plugins extend it with tasks and triggers for databases, cloud providers and AI models."
        metadata:
          doc: "kestra-intro"
    metadata:
      source: "splitter-run"
      team: "data"
  - id: search
    type: io.kestra.plugin.ai.rag.Search
    provider:
      type: io.kestra.plugin.ai.provider.DockerModel
      modelName: ai/nomic-embed-text-v1.5
      baseUrl: http://172.17.0.1:12434/engines/v1
    embeddings:
      type: io.kestra.plugin.ai.embeddings.KestraKVStore
      kvName: qa_search
    query: "What is Kestra?"
    maxResults: 3
    minScore: 0.0
  - id: read_kv
    type: io.kestra.plugin.core.kv.Get
    namespace: qa.rag
    key: qa_search
```
</details>

**Gantt** ([screenshot](https://claude.ai/code/artifact/3f56a295-03df-45ba-a9c5-ea3e853b9357#ingest_with_splitter_and_search-gantt))
| Task | Status | Duration |
|------|--------|----------|
| `ingest` | SUCCESS | 0.15s |
| `search` | SUCCESS | 0.11s |
| `read_kv` | SUCCESS | 0.04s |
| **Total** | **SUCCESS** | **0.40s** |

**Stored metadata** ([outputs screenshot](https://claude.ai/code/artifact/3f56a295-03df-45ba-a9c5-ea3e853b9357#ingest_with_splitter_and_search-outputs))
```json
{"doc": "kestra-intro", "index": "0", "source": "splitter-run", "team": "data"}
```
Metadata lands on the split segment, confirming it is attached before `ingestor.ingest(batch)`. `rag.Search` still returns results — the retrieval path is unaffected.

---

### Out-of-scope findings (pre-existing on `main`, not caused by this PR)

Both were hit while writing the QA flows. Neither is touched by this diff; filing them is a separate call.

1. **`documentSplitter` NPEs when `maxOverlapSizeInChars` is omitted.** `IngestDocument.DocumentSplitter.maxOverlapSizeInChars` is annotated `@NotNull` but has no default and is not validated, so a splitter without it fails at runtime with:
   ```
   Cannot invoke "java.lang.Integer.intValue()" because the return value of
   "io.kestra.plugin.ai.rag.IngestDocument$DocumentSplitter.getMaxOverlapSizeInChars()" is null
   ```
   A `@Builder.Default` of `0` (or enforced validation) would turn this into either a working default or a clear config error.

2. **`rag.Search.minScore` is required with no default.** Omitting it rejects the flow at save time with `search.minScore: must not be null`, which is a rough first-use experience for a property most users have no opinion about.

### Timeout notes

None. Every scenario completed well inside the 30s budget.

src/main/java/io/kestra/plugin/ai/rag/IngestDocument.java
```diff
@@ -265,6 +309,44 @@ public Output run(RunContext runContext) throws Exception {
...
+            return Map.of();
+        }
+
+        var validated = new LinkedHashMap<String, Object>();
+        for (var entry : rendered.entrySet()) {
+            var value = entry.getValue();
+            if (value == null) {
+                continue;
+            }
+            if (!SUPPORTED_METADATA_TYPES.contains(value.getClass())) {
```
jymaire, line 348: 🟡 [Guidelines/UX] Boolean metadata is a plausible YAML value that now hard-fails
Problem: a user writing `metadata:` / `  public: true` gets a `Boolean` out of `asMap(String, Object)`, which is not in `SUPPORTED_METADATA_TYPES`, so the whole ingestion aborts before it starts. Same for `BigInteger`/`BigDecimal` if a YAML number exceeds `Long`/`Double` precision. The message is clear and names the key, so this is not a blocker — but rejecting a boolean outright is harsher than needed when `String.valueOf(value)` would round-trip fine.
Fix: either coerce `Boolean` (and `BigInteger`/`BigDecimal`) to `String` before validating, or keep rejecting and add "booleans are not supported — quote them" to the `@Schema` description so the constraint is discoverable before the run fails.

src/main/java/io/kestra/plugin/ai/rag/IngestDocument.java
```diff
@@ -265,6 +309,44 @@ public Output run(RunContext runContext) throws Exception {
             .build();
     }
 
+    /** Adds the top-level metadata to a document, keeping any value already set by the loader or by the inline document. */
+    private static void applyMetadata(Document document, Map<String, Object> base) {
```
jymaire, line 319: 🟢 [Performance] Allocation per document when metadata is unset
Problem: `applyMetadata` allocates a `LinkedHashMap` for every document even when `base` is empty, which is the common case. On a 100k-document ingest that is 100k throwaway maps — negligible next to the embedding calls, but free to avoid.
Fix: `if (base.isEmpty()) { return; }` as the first statement.

src/main/java/io/kestra/plugin/ai/rag/IngestDocument.java
```diff
@@ -224,7 +264,9 @@ public Output run(RunContext runContext) throws Exception {
         for (String uri : runContext.render(fromInternalURIs).asList(String.class)) {
             try (InputStream file = runContext.storage().getFile(URI.create(uri))) {
                 byte[] bytes = file.readAllBytes();
-                batch.add(Document.from(new String(bytes)));
+                var doc = Document.from(new String(bytes));
```
jymaire: 🟢 [Guidelines] Platform-default charset
Problem: `new String(bytes)` decodes with the JVM default charset. Pre-existing, but this line is rewritten by the diff so it is worth fixing while here — a non-UTF-8 default would mangle the very accented content the tests ingest (`I'm Loïc`).
Fix: `new String(bytes, StandardCharsets.UTF_8)`.

src/test/java/io/kestra/plugin/ai/rag/IngestDocumentTest.java
```diff
@@ -221,6 +223,171 @@ void inlineDocumentsWithBulkSize() throws Exception {
...
+        assertThat(metadata.getFirst()).containsEntry("source", "manual-run").containsEntry("team", "data");
+    }
+
+    @Test
+    void topLevelMetadataFromExternalURLs() throws Exception {
+        RunContext runContext = ollamaRunContext();
+
+        var task = ingestTaskBuilder()
+            .metadata(Property.ofValue(Map.of("source", "manual-run")))
+            .fromExternalURLs(Property.ofValue(List.of("https://dummyjson.com/products/1")))
```
jymaire: 🟢 [Tests] Second live-internet dependency in CI
Problem: this adds a second test that depends on `dummyjson.com` being reachable. Consistent with the existing test at line 172, so not a regression, but the assertion here is purely about metadata propagation — the external fetch adds flakiness surface for nothing.
Fix: optional — cover the `fromExternalURLs` metadata path against a WireMock stub, or fold the metadata assertion into the existing external-URL test.

jymaire (commented):
## Kestra Plugin Code Review

### Business Requirements — met
Issue #395 asked to either apply the top-level `metadata` to every source or remove the property. The PR takes the first option and covers all four sources (`fromPath`, `fromInternalURIs`, `fromExternalURLs`, `fromDocuments`). No scope creep.

### Kestra Guidelines — 4 findings

**🟠 Package doc not updated** — `src/main/resources/doc/io.kestra.plugin.ai.md:31` enumerates `IngestDocument`'s properties (`fromPath`, `fromInternalURIs`, `fromExternalURLs`, `fromDocuments`, `documentSplitter`, `drop`, `bulkSize`) and mentions `metadata` only as an inline-document field. Top-level `metadata` is now a real capability and is absent from the plugin overview page — the one place a user looks before reading the property schema. Add a sentence to that paragraph: top-level `metadata` applies to every document from every source, existing keys win, and which value types are supported.

**🟡 Behavior change worth a release note** — a flow that today sets an unsupported top-level `metadata` value (a list, a nested map, a boolean) succeeds, because the property is silently ignored. After this PR the same flow fails at task start. That is the correct behaviour and it fails loudly rather than silently, but it is a user-visible break for existing flows and is not mentioned in the PR description. Worth one line in the release notes.

🟡 `IngestDocument.java:338` — boolean / big-number metadata rejected (inline comment).
🟢 `IngestDocument.java:267` — platform-default charset on a rewritten line (inline comment).

Verified correct, no action needed:
- `Property<T>` usage, `@PluginProperty(group = "advanced")`, multiline `@Schema`, and the new `@Example` (`full = true`, flow id, namespace, no `lang`, `{{ secret(...) }}`) all match conventions.
- `validateMetadata` runs at line 211, **before** `provider.embeddingModel()` and before `embeddings.embeddingStore(..., drop)` — so an invalid value cannot drop an existing store or leave a partial write. This ordering is load-bearing and correct.
- `SUPPORTED_METADATA_TYPES` exactly matches the boxed entries of LangChain4j `Metadata.SUPPORTED_VALUE_TYPES` (verified against `langchain4j-core-1.19.0` bytecode), and `Metadata.putAll` re-validates against that same set — so nothing accepted by `validateMetadata` can throw inside `applyMetadata`. Skipping `null` also matches `Metadata.validate`, which rejects nulls. The two helpers agree.
- `applyMetadata` is called before `batch.add` at all four sites, so metadata is attached before any `flushBatch` boundary and before the splitter runs — segments inherit it via `metadata().copy()`.
- Type widening `Map<String, String>` → `Map<String, Object>` is deserialization-compatible (existing string-valued YAML still binds) and aligns with `InlineDocument.metadata`, which was already `Map<String, Object>`. No polymorphic deserialization is introduced.

### Security (OWASP Top 10:2025 + KPS) — 0 findings
Metadata is operator-authored flow config, not attacker-controlled input, and is never logged. The exception message echoes the key name and the value's `getSimpleName()`, never the value itself — no leak via A10. No secrets, no new dependency, no HTTP client change, no deserialization surface (values bind as untyped `Object` with no `@JsonTypeInfo`). Metadata size is bounded by the flow definition, so KPS-08 does not apply.

### Performance — 1 finding
🟢 `IngestDocument.java:313` — one throwaway `LinkedHashMap` per document even when `metadata` is unset (inline comment). Everything else is O(docs × keys) with tiny constants, dwarfed by the embedding calls.

### Tests
Assertions are meaningful, not tautological: they read back the actual KV-store payload and assert on the stored `metadata` node, and the precedence tests would fail under a naive `putAll` overwrite. The no-metadata regression guard asserts `doesNotContainKey("source")`, which is the right negative. Two nits: a second live-internet dependency (inline comment), and the new `ingestTaskBuilder()` helper is not applied to the five pre-existing tests, which still duplicate the builder verbatim — fine to leave out of this diff, but the file is now half-refactored.

**Verdict:** REQUEST CHANGES — only the package doc needs to land before merge; everything else is optional.

src/main/java/io/kestra/plugin/ai/rag/IngestDocument.java
jymaire: Kept the fail-fast behaviour rather than coercing to String: silently turning `true` into `"true"` would make the stored metadata type depend on what LangChain4j happens to accept, and a later widening of the supported types would then be a silent behaviour change for anyone relying on the coerced form. Instead the constraint is now discoverable before runtime — the `@Schema` description spells out that a boolean, a list, a map, or an out-of-range number is rejected before any document is ingested, and points at quoting the value; the thrown message carries the same hint. Also flagged in the PR description as a user-visible behaviour change worth a release note.

src/main/java/io/kestra/plugin/ai/rag/IngestDocument.java
jymaire: Added the early `if (base.isEmpty()) return;` guard in `applyMetadata`.

src/main/java/io/kestra/plugin/ai/rag/IngestDocument.java
jymaire: Fixed: `new String(bytes, StandardCharsets.UTF_8)`.

src/test/java/io/kestra/plugin/ai/rag/IngestDocumentTest.java
jymaire: The new test now uses a `file:` URL built from a working-directory file instead of `dummyjson.com`. It still goes through `UrlDocumentLoader`, and since that loader injects the `url` metadata key the test additionally asserts the top-level `metadata` does not overwrite it. The pre-existing `externalURLs()` test at line 172 still hits the network — out of scope here. Also routed the five pre-existing tests through `ingestTaskBuilder()` / `ollamaRunContext()`.

jymaire (commented):
## Kestra Plugin Code Review — cycle 2

Re-review of the fix batch (`6c8fbb3..806b588`, 3 files, +31/−95) in the context of the full PR diff. All 4 findings from [cycle 1](https://github.com/kestra-io/plugin-ai/pull/417#pullrequestreview-5110064302) are addressed.

### Business Requirements — met
Top-level `metadata` now reaches all four sources, existing keys win on collision, and the previously misleading `@Schema` note is replaced by an accurate description.

### Kestra Guidelines — 0 new findings

Resolution of the cycle-1 findings:

1. **Doc (blocking) — resolved.** `io.kestra.plugin.ai.md:31` now covers the behaviour inline. I checked the claim against the LangChain4j bytecode: `FileSystemSource.metadata()` injects exactly `file_name` + `absolute_directory_path`, `UrlSource.metadata()` injects exactly `url` — the three keys the sentence lists. Accurate, and matching the page's one-paragraph-per-task style is the right call over a new section. The sentence omits the fail-fast rejection, but the property `@Schema` carries it and that is what the UI renders, so this is not a gap worth another cycle.

2. **`Boolean` / `BigInteger` / `BigDecimal` — accepting fail-fast, dropping the coercion request.** The reasoning holds: coercing `true` → `"true"` silently picks a storage type on the user's behalf, and if the accepted set widens later those users would flip from `String` to `Boolean` in the store, breaking metadata filters with no diff to point at. Fail-fast keeps the stored type a user decision. The constraint is now statically discoverable in the `@Schema` and the thrown message names the key, the offending type, the valid set, and the fix ("quote the value to send it as a String") — that satisfies the actionable-error-message bar. No further change wanted here.

3. **`applyMetadata` allocation — resolved.** `if (base.isEmpty()) return;` short-circuits before the `LinkedHashMap`. Semantically identical to the previous path (empty `missing` → skipped `putAll`).

4. **Charset — resolved.** `new String(bytes, StandardCharsets.UTF_8)`. A no-op under JEP 400 but correct on an overridden default; import added and used.

### Test refactor — no coverage lost (verified against `main`)

I diffed the five pre-existing tests against `main:src/test/java/io/kestra/plugin/ai/rag/IngestDocumentTest.java`. Each asserted exactly two things before — `getIngestedDocuments()` and `assertKvStore(kvStore, kvKey, n)` — and each still asserts both with the same expected counts (1, 1, 2, 2, 12). The −95 lines are entirely the inlined `Ollama` provider / `KestraKVStore` / `drop(true)` builder and the `runContextFactory.of(...)` block, now in `ollamaRunContext()` / `ingestTaskBuilder()`, which reproduce them field for field. Pure boilerplate removal.

**The `file:` URL swap genuinely still exercises `UrlDocumentLoader`.** Confirmed at the bytecode level: production calls `UrlDocumentLoader.load(String, DocumentParser)`, which is the exact overload the test drives; it routes `createUrl` → `new URL(..)` → `UrlSource`, whose `metadata()` is `Metadata.from("url", url.toString())`. Only the JDK stream handler differs (file vs http) — the document-loading and metadata-injection path is identical. The round-trip is stable, so the assertion is not flaky:

```
built = file:/home/.../document.txt
rtrip = file:/home/.../document.txt   (equal = true)
```

The one-line comment explaining the `file:///x` → `file:/x` normalization documents a genuinely non-obvious WHY and is the kind of comment the guidelines permit. Net, this test is *stronger* than before: it drops a network dependency and adds a loader-key collision assertion (`url: "should-not-win"` must not overwrite the loader value) that did not exist in cycle 1.

### Security (OWASP Top 10:2025 + KPS) — 0 findings
Fix batch touches a charset, a guard clause, two doc strings, and test scaffolding. Nothing disturbs the cycle-1 security clearance. No secrets, no new I/O surface, no new dependency.

### Performance — 0 findings
The `base.isEmpty()` guard removes a per-document allocation on the (common) no-metadata path. No new allocation, loop, or blocking call introduced.

### Non-blocking observation
`externalURLs()` still reaches `https://dummyjson.com` and remains a live-network dependency in CI. Correctly scoped out of this PR — worth a follow-up applying the same `file:`-URL or WireMock treatment now that the pattern exists.

**Verdict:** APPROVE — no blocking or change-requesting findings. Posted as a comment review because GitHub does not allow approving your own PR; treat this as the review gate passing.

All four cycle-1 threads are correctly resolved. Good to ship.

src/main/java/io/kestra/plugin/ai/rag/IngestDocument.java
```diff
@@ -112,6 +140,9 @@
     aliases = "io.kestra.plugin.langchain4j.rag.IngestDocument"
 )
 public class IngestDocument extends Task implements RunnableTask<IngestDocument.Output> {
+    private static final Set<Class<?>> SUPPORTED_METADATA_TYPES =
```
fdelbrayelle, line 143: 🟡 [Guidelines] Documented types partly unreachable via YAML
Problem: `UUID` and `Float` are listed as supported metadata value types (mirroring langchain4j's `Metadata.SUPPORTED_VALUE_TYPES`), but `Property.asMap`'s Jackson-based deserialization of rendered YAML scalars can only ever produce `String`, `Integer`, `Long`, or `Double` — never `UUID` or `Float`. Not a bug (accepted superset matching the SDK contract), but a user following the docs literally with a UUID-looking string will just get a `String`, not a rejection or special handling.
Fix: Either add a test constructing the task with `Property.ofValue(Map.of("id", UUID.randomUUID()))` (bypasses rendering, preserves the exact type) to prove the two extra types are honored when reachable, or note in the `@Schema` that `UUID`/`Float` only apply when set programmatically.

src/main/java/io/kestra/plugin/ai/rag/IngestDocument.java
```diff
@@ -265,6 +312,48 @@ public Output run(RunContext runContext) throws Exception {
...
+        });
+
+        if (!missing.isEmpty()) {
+            document.metadata().putAll(missing);
+        }
+    }
+
+    /** Fails the whole task before any ingestion happens, so an unsupported value cannot leave a partially ingested store. */
+    private static Map<String, Object> validateMetadata(Map<String, Object> rendered) {
+        if (rendered == null || rendered.isEmpty()) {
```
fdelbrayelle: 🟢 [Guidelines] Dead null check
Problem: `rendered == null` is unreachable — `RunContextProperty#asMap` is documented to return an empty map, never null, for a null property.
Fix: Simplify to `if (rendered.isEmpty()) { return Map.of(); }` (non-blocking).

src/main/java/io/kestra/plugin/ai/rag/IngestDocument.java
```diff
@@ -224,7 +267,9 @@ public Output run(RunContext runContext) throws Exception {
         for (String uri : runContext.render(fromInternalURIs).asList(String.class)) {
             try (InputStream file = runContext.storage().getFile(URI.create(uri))) {
                 byte[] bytes = file.readAllBytes();
-                batch.add(Document.from(new String(bytes)));
+                var doc = Document.from(new String(bytes, StandardCharsets.UTF_8));
```
fdelbrayelle, line 273: 🟡 [Guidelines] Undocumented scope-creep behaviour change
Problem: switching from `new String(bytes)` (platform default charset) to explicit UTF-8 is a good fix but is unrelated to the ticket and changes decoding behaviour on non-UTF-8-default JVMs. It isn't mentioned in the PR's "user-visible behaviour change" section.
Fix: Non-blocking — just add a one-line mention alongside the existing release note, or split into its own commit/PR next time to keep the diff scoped to the metadata fix.

src/test/java/io/kestra/plugin/ai/rag/IngestDocumentTest.java
```diff
@@ -221,6 +145,178 @@ void inlineDocumentsWithBulkSize() throws Exception {
...
+        Optional<KVValue> kvValue = kvStore.getValue(kvEntry.get().key());
+        JsonNode jsonNode = JacksonMapper.ofJson().readTree(kvValue.orElseThrow().value().toString());
+
+        List<Map<String, Object>> metadata = new ArrayList<>();
+        for (JsonNode entry : jsonNode.get("entries")) {
+            JsonNode node = entry.get("embedded").get("metadata");
+            if (node.has("metadata")) {
+                node = node.get("metadata");
+            }
+            metadata.add(JacksonMapper.ofJson().convertValue(node, Map.class));
```
fdelbrayelle: 🟢 [Guidelines] Raw type + duplicated KV-parsing logic
Problem: `convertValue(node, Map.class)` uses a raw `Map` type (unchecked warning when added to `List<Map<String,Object>>`), and this KV-store JSON parsing duplicates what `assertKvStore` already does a few lines below.
Fix: Use `new TypeReference<Map<String,Object>>(){}` and consider extracting a shared `readKvJson(...)` helper. Test-only, non-blocking.

fdelbrayelle (approved):
## Kestra Plugin Code Review

### Business Requirements — met
Issue #395: top-level `metadata` on `rag.IngestDocument` was declared but never read, silently dropping user input and leaving `fromPath`/`fromInternalURIs`/`fromExternalURLs` with no way to attach metadata at all. The PR renders `metadata` once, merges it into every document from all four sources at `batch.add` time (so a configured `documentSplitter` propagates it to each segment), and preserves existing-key precedence (inline `fromDocuments[].metadata` and loader-injected keys like `file_name`/`absolute_directory_path` always win). It chose "apply the property" over "remove it", one of the two acceptable resolutions named in the issue. The intentional behaviour change (an unsupported value type now fails fast instead of being silently accepted) is explicitly called out in the PR description as a release note — good transparency.

### Kestra Guidelines — 0 blocking findings
- Rendered value correctly prefixed `rMetadata`; rendered once at the top of `run()` and reused, avoiding repeated Pebble evaluation per document.
- `@Schema` description rewritten with an accurate, multi-line `"""` block; new `@Example` is `full = true`, has a flow id/namespace, and uses `"{{ secret('GEMINI_API_KEY') }}"` for the credential — closing `"""` aligned with the example's `id:` line as required.
- Package how-to doc (`doc/io.kestra.plugin.ai.md`) updated to describe the now-functional property, including the precedence rule.
- The comments on `applyMetadata`/`validateMetadata` document non-obvious invariants (precedence order, atomicity of the pre-flight check) in one line each — compliant with the "WHY, not WHAT" comment rule.
- `SUPPORTED_METADATA_TYPES` in the plugin exactly mirrors langchain4j's own `Metadata.SUPPORTED_VALUE_TYPES` (verified by decompiling `langchain4j-core`), so the pre-flight check fails fast with a message naming the offending key instead of surfacing an opaque exception mid-ingestion from `Metadata.putAll` after some batches may already have been flushed to the store.
- Good accompanying DRY cleanup of the test class: five near-identical `IngestDocument.builder()...` blocks collapsed into shared `ollamaRunContext()` / `ingestTaskBuilder()` helpers.

Minor, non-blocking points (see inline comments):
- 🟡 Documented supported metadata types (`UUID`, `Float`) are largely unreachable via normal YAML/Pebble rendering.
- 🟢 Dead `rendered == null` check in `validateMetadata`.
- 🟡 Unannounced charset fix (`fromInternalURIs`) is minor scope creep, not mentioned in the release note.
- 🟢 Test helper duplication / raw `Map` type in `storedMetadata`.

### Security (OWASP Top 10:2025 + KPS) — 0 issues found
- No secrets or new external calls involved; the validation error message intentionally omits the metadata value itself (only the key name and the value's simple class name), which is more conservative than langchain4j's own exception message and avoids leaking potentially sensitive metadata content into logs/execution errors.
- No new HTTP/deserialization surface introduced.

### Performance — 0 issues found
- `applyMetadata` early-returns when the top-level map is empty, so the common case (no top-level metadata) adds no per-document allocation.
- Per-document `LinkedHashMap` allocation in `applyMetadata` is bounded by the (small, user-configured) metadata key count, not by input size.

**Verdict:** APPROVE

src/main/java/io/kestra/plugin/ai/rag/IngestDocument.java
jymaire: Took the documentation option rather than the extra test. Added to the `@Schema`:

> A value written in a flow can only ever be a String, an Integer, a Long or a Double: UUID and Float are accepted for parity with the underlying model and are only reachable when the property is set programmatically.

A `Property.ofValue(Map.of("id", UUID.randomUUID()))` test would only assert that `SUPPORTED_METADATA_TYPES` contains `UUID.class`, which is a restatement of the constant rather than a behaviour a user can reach — so the constraint is now discoverable in the UI-rendered schema instead. Kept the accepted set mirroring `Metadata.SUPPORTED_VALUE_TYPES` so the pre-flight check cannot reject something the store would have taken.

src/main/java/io/kestra/plugin/ai/rag/IngestDocument.java
jymaire: Fixed — `validateMetadata` now starts with `if (rendered.isEmpty()) { return Map.of(); }`.

src/main/java/io/kestra/plugin/ai/rag/IngestDocument.java
jymaire: Fair — added a paragraph to the release-note section of the PR description rather than reverting it, since the line is inside the diff anyway:

> Also in this PR, `fromInternalURIs` now decodes documents as explicit UTF-8 instead of the JVM default charset. This is a no-op on a JVM whose default is already UTF-8 (the default since JEP 400) but changes decoding on one started with an overridden `-Dfile.encoding`.

Point taken on keeping the diff scoped next time.

src/test/java/io/kestra/plugin/ai/rag/IngestDocumentTest.java
jymaire: Fixed both. `readKvJson(kvStore, kvKey)` now does the entry/value lookup and `readTree`, and is shared by `storedMetadata` and `assertKvStore`; the metadata node is read through a `private static final TypeReference<Map<String, Object>> METADATA_TYPE`, so no more raw type or unchecked warning.

